PR #39, opened 2015-06-28
Retroactive fix to old par contract calculation in rare cases

sorenhein:
Around v2.8.1 it wad discovered that the old par contract calculation could incomplete results in certain  cases.  As v2.5.2 was an important release and it affected about 4 hands in 1000 in this release, we decided to fix this.


sorenhein:
Was already merged elsewhere, I think.

PR #41, opened 2015-07-31
Allow cross compile to windows

suokko:
There is fairly simple change to mingw makefile that would allow cross-compiling from linux to windows. I don't have access to a windows machine in my home so I can't test if msys/mingw tool chain has windres  prefixed with complete tool chain name.

I use cross-compiling and wine to check that dealer code runs in windows. I recently added code to dynamicaly load libdds.so/dds.dll.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.o
 *.a
 *.dll
+*.so
 *.obj
 *.exe
 *.def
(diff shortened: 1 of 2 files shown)


sorenhein:
On Windows there is only windres.  I can't think of a simple way to accommodate both.

How would you feel about aliasing windres to your specific windres?  That should fool the Makefile?

sorenhein:
This looks different v.2.9.0, so I'm sorry I can't take the merge exactly as it is.  But thanks!

PR #84, opened 2018-09-24
Fix compiler warnings on platforms where char is unsigned

df7cb:
If char is unsigned, there were two warnings that made the build fail
because of `-Werror=sign-promo`:

Init.cpp: In function 'void SetDealTables(ThreadData*)':
Init.cpp:534:9: error: conversion to 'signed char' from 'char' may change the sign of the result [-Werror=sign-conversion]
         static_cast<char>(handLookup[s][topBitNo]);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

TransTableL.cpp: In member function ‘void TransTableL::PrintNodeValues(std::ofstream&, const nodeCardsType&) const’:
TransTableL.cpp:1028:24: error: passing ‘char’ chooses ‘int’ over ‘unsigned int’ [-Werror=sign-promo]
     to_string(np.lbound) << " to " <<
                        ^

Seen on Debian's buildds: https://buildd.debian.org/status/fetch.php?pkg=dds&arch=ppc64el&ver=2.9.0-2&stamp=1537788660&raw=0

### DIFF
--- a/src/TransTableL.cpp
+++ b/src/TransTableL.cpp
@@ -1025,8 +1025,8 @@ void TransTableL::PrintNodeValues(
     cardSuit[3] << cardRank[15-static_cast<int>(np.leastWin[3])] << "\n";
 
   fout << setw(16) << left << "Bounds" << 
-    to_string(np.lbound) << " to " <<
-    to_string(np.ubound) << " tricks\n";
+    to_string(static_cast<int>(np.lbound)) << " to " <<
+    to_string(static_cast<int>(np.ubound)) << " tricks\n";
 
   fout << setw(16) << left << "Best move" <<
     cardSuit[ static_cast<int>(np.bestMoveSuit) ] <<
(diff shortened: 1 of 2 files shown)


src/TransTableL.cpp
```diff
@@ -1025,8 +1025,8 @@ void TransTableL::PrintNodeValues(
     cardSuit[3] << cardRank[15-static_cast<int>(np.leastWin[3])] << "\n";
 
   fout << setw(16) << left << "Bounds" << 
-    to_string(np.lbound) << " to " <<
-    to_string(np.ubound) << " tricks\n";
+    to_string(static_cast<int>(np.lbound)) << " to " <<
```
suokko, line 1028: Why is unsigned char generate warning when calling to_string?

suokko: I was stupid. Explanation was in the commit message. The github provided horrible formating didn't do any favors for reading the compiler warning.

sorenhein:
Thanks

PR #85, opened 2018-09-24
Drop -mtune=generic

df7cb:
"-mtune=generic" works on x86 only and fails on all other architectures.

Failing build logs: https://buildd.debian.org/status/logs.php?pkg=dds&ver=2.9.0-1


sorenhein:
Thanks

PR #86, opened 2018-09-24
In C mode, include <stdbool.h>

df7cb:
When compiling with a C compiler, "bool" is not defined, so include
<stdbool.h> to make it available.


rimmington (approved):

sorenhein:
Thanks

PR #88, opened 2018-10-10
Linux memory

df7cb:

PR #89, opened 2018-12-04
Avoid nullptr access

ivannp:
Merely loading libdds.so on Ubuntu crashes the parent program on exit. To repro, run the following python command:

    python -c 'from ctypes import *; dll = CDLL("libdds.so");'

I chased it down to trying to release memory on exit. It looks like the logic behind freeing the memory is broken, when the thread count is less than what's in the array. One might argue that this fix fixes the symptom, but it's better to avoid accessing pointers blindly.

This fix fixes the problem only on 18.04. On 16.04 it seems we are hitting the following line from Memory.cpp:

    cout << "Memory::GetPtr: " << thrId << " vs. " << nThreads << endl;

From a quick inspection of the code, I expect to be similar issue. Haven't spent the time debugging it yet.

### DIFF
--- a/src/Memory.cpp
+++ b/src/Memory.cpp
@@ -37,7 +37,7 @@ void Memory::ResetThread(const unsigned thrId)
 
 void Memory::ReturnThread(const unsigned thrId)
 {
-  if(memory[thrId] == nullptr) return;
+  if(memory[thrId] == nullptr || memory[thrId]->transTable == nullptr) return;
   memory[thrId]->transTable->ReturnAllMemory();
   memory[thrId]->memUsed = Memory::MemoryInUseMB(thrId);
 }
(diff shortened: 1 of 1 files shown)


t3nsor:
I think the code that determines the amount of free memory (Memory.cpp:228) may be the issue. On my system (Debian 9) it produces 0. Note that the full output of `free -k` for me is:

              total        used        free      shared  buff/cache   available
Mem:       16198028     5259516      808524      173588    10129988    10428852
Swap:             0           0           0

The output format was changed about 4 years ago: https://gitlab.com/procps-ng/procps/commit/f47001c9e91a1e9b12db4497051a212cf49a87b1

df7cb:
See #88.

sorenhein:
Thanks. It did fix your issue, but it seems that rimmington fixed some larger issues that included yours, so I've gone with his pull request.

PR #93, opened 2019-04-20
Three small fixes

elkhrt:
1. Remove File from the global namespace. Ideally all the code would be in a namespace, but File is the only one that caused a conflict for me.
2. Fix a reference-before-initialization issue (innocuous in practice).
3. Include-what-we-use


sorenhein:
Thanks

PR #97, opened 2019-11-04
one .travis.yml example to build on ubuntu

### LINKED ISSUE #94: sample to generate linux binary using .travis.yml
it will be nice to have automation inside the repo to generate binary automatically like using ` .travis-ci.yaml`, therefore it can also be a binary release.

Or someone had clear instruction on how to build binary from base ubuntu 18.04

As an end-user, I want to have a linux binary to analyse my hand card.

larrycai:
resolves #94 
* add one sample `.travis.yml` to start with for ubuntu env
* fix one bug in `Makefile_linux_shared`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ addons:
     - libomp-dev 
 script:
   - cd src
-  - pwd
-  - ls
+  - make Makefiles/Makefile_linux_shared
   - make install -f Makefiles/Makefile_linux_shared
   - cd ../example
   - export LD_LIBRARY_PATH=$PWD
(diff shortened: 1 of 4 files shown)


larrycai:
With Antony's patch from [anntzer/redeal](https://github.com/anntzer/redeal), it works finally, sees https://travis-ci.com/larrycai/dds/jobs/254796009

* use `grep /proc/meminfo` for memory check instead `free` command in modern unix system
* delete `FreeMemory()`

check codes for detail, it shall work for other unix as well now

zzcgumn:
These crashes have, to my knowledge, been resolved in version 3.0.0. There are also GitHub workflows that test the library on Linux, macOS and windows. Is there still a need to add support for Travis CI?

PR #98, opened 2020-01-07
Avoid OOB memory access + shrink Memory members

### LINKED ISSUE #91: Segmentation fault when libdds.so is unload
I compiled libdds.so on Linux with the default parameter and I'm trying to use it with ctypes in Python. I noticed, when I unload the library at the end of a Python script, I get a segmentation fault.

Here is a toy example of Python script:
```
from ctypes import *
dll = cdll.LoadLibrary("libdds.so")
```
When I run it:
```
$ python libdds.py
Segmentation fault (core dumped)
```

According to the debugger, the issue come from FreeMemory function:
```
$ gdb
(gdb) file python
Reading symbols from python...done.
(gdb) run libdds.py
Starting program: /home/pierre/anaconda3/bin/python libdds.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff5e59454 in Memory::ReturnThread(unsigned int) ()
   from /home/pierre/Documents/python/bridge/python-dds-master_copy/libdds.so
(gdb) bt
#0  0x00007ffff5e59454 in Memory::ReturnThread(unsigned int) ()
   from /home/pierre/Documents/python/bridge/python-dds-master_copy/libdds.so
#1  0x00007ffff5e5cb31 in FreeMemory () from /home/pierre/Documents/python/bridge/python-dds-master_copy/libdds.so
#2  0x00007ffff7de7de7 in _dl_fini () at dl-fini.c:235
#3  0x00007ffff6a0cff8 in __run_exit_handlers (status=0, listp=0x7ffff6d975f8 <__exit_funcs>, 
    run_list_atexit=run_list_atexit@entry=true) at exit.c:82
#4  0x00007ffff6a0d045 in __GI_exit (status=<optimized out>) at exit.c:104
#5  0x00007ffff69f3837 in __libc_start_main (main=0x400ab0 <main>, argc=2, argv=0x7fffffffde68, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffde58) at ../csu/libc-start.c:325
#6  0x00000000004009e9 in _start ()
```

An idea about how to correct that issue?

rimmington:
Includes some defensive tweaks I applied while tracking down the cause of a segfault.

Fixes #91. Fixes the root cause of #89.

### DIFF
--- a/src/Init.cpp
+++ b/src/Init.cpp
@@ -28,8 +28,6 @@ void InitConstants();
 
 void InitDebugFiles();
 
-void FreeThreadMem();
-
 
 int lho[DDS_HANDS] = { 1, 2, 3, 0 };
 int rho[DDS_HANDS] = { 3, 0, 1, 2 };
@@ -594,13 +592,6 @@ void ResetBestMoves(
 void STDCALL GetDDSInfo(DDSInfo * info)
 {
   (void) sysdep.str(info);
-}
-
-
-void FreeThreadMem()
-{
-  for (unsigned thrId = 0; thrId < sysdep.NumThreads(); thrId++)
-    memory.ResetThread(thrId);
 }
 
 
--- a/src/Init.cpp
+++ b/src/Init.cpp
@@ -350,7 +350,7 @@ void InitConstants()
 
 void InitDebugFiles()
 {
-  for (unsigned thrId = 0; thrId < sysdep.NumThreads(); thrId++)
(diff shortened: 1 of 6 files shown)


sorenhein:
Thanks! Will be added to the next release.

PR #101, opened 2020-01-12
New DDS_NO_DUMP_ON_ERROR flag to skip dump.txt generation

rimmington:
It's not always useful to produce these files, so allow disabling their generation at compile time. This is a `NO` flag to avoid changing default behaviour.


sorenhein:
Thanks

PR #102, opened 2020-01-18
Export SolveAllBoardsBin + update documentation

rimmington:
Adds a binary equivalent of `SolveAllBoards`.

An examination of the source reveals that the `chunkSize` parameters are no longer used, and the `SolveAllChunks` functions are marked as "Historical aliases". The documentation is updated to match.


sorenhein:
Thanks

PR #108, opened 2020-06-07
Fix crash when unloading the library

suokko:
I noticed dds was causing crash in exit because libEnd was called after memory::~memory(). Static destructor are executed in unspecified order. To avoid order issues I decided to make changes which moves all memory cleanup code to automatically generated destructors using stl and std::unique_ptr memory management.

I noticed there is pull request #90. But it seems to have issues still so I decided to provide alternative attempt at fixing same issue.

### DIFF
--- a/src/Memory.cpp
+++ b/src/Memory.cpp
@@ -9,6 +9,7 @@
 
 
 #include "Memory.h"
+#include "Init.h"
 
 
 Memory::Memory()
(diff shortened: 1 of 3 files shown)


zzcgumn:
These crashes should have been fixed in version 3.0.0. Would be appreciated if you can confirm.

PR #109, opened 2020-06-07
Make sure link time optimization sees same compiler flags

suokko:
gcc prefers that link time optimization flags must be matching
for compile and link commands.

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

"To use the link-time optimizer, -flto and optimization options
should be specified at compile time and during the final link."


sorenhein:
Thanks

PR #113, opened 2020-12-16
Build on FreeBSD

aft90:
Hi! Thanks for creating this library. I found myself needing to compile it on FreeBSD (with clang and BSD make which are the default on that platform) so I thought I'd share the changes I had to make:

* added makefiles for the main library (static & shared), test and examples
* added `stdio.h` to `Par.cpp` for `snprintf`
* remove the `-Wlogical-op`, `-Wnoexcept` and `Wstrict-null-sentinel` flags because clang 10 on FreeBSD doesn't recognise them
* turn off some more warnings in `System.cpp` caused by clang with boost
* added checks for the `__FreeBSD__` compiler macro where appropriate

### DIFF
--- a/examples/AnalyseAllPlaysBin.cpp
+++ b/examples/AnalyseAllPlaysBin.cpp
@@ -30,7 +30,7 @@ int main()
   char line[80];
   bool match;
 
-#if defined(__linux) || defined(__APPLE__)
+#if defined(__linux) || defined(__APPLE__) || defined(__FreeBSD__)
   SetMaxThreads(0);
 #endif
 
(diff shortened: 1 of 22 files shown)


zzcgumn:
Version 3.0.0 builds with the bazel clang toolchain. I have not added explicit support for FreeBSD but am happy to include it. Would it be possible for you to open a pull request with a FreeBSD workflow?

aft90:
Wow, this is a blast from the past! I'll see what I can do. I'm not familiar with Bazel and I've had 2 children in the mean time - but if time permits I'll give it a try.

PR #115, opened 2021-02-26
Better reuse of the transposition table etc.

### LINKED ISSUE #114: Better reuse of the transposition table?
A normal usecase for the DDS SolveBoard() method is to:
  1) Call SolveBoard() for the opening lead and play the suggested card for say West.
  2) Have the user (North)  play a card.
  3) Call SolveBoard() again 
  etc.

As I see it, it must be possible to reuse the transposition table and the rel/aggr structures in all calls after 1).
To make this possible, the test for NewDeal should only validate that all cards in the current deal was in the same hand as when the transposition table was initiated.  

I've actually got it working with very minimal changes to the code. That is when the new deal is a subset of the old one. But is now working on a superset version also. By this I mean situations where all cards in the deal stored in the transpositiontable are placed in the same hands in the new deal and none of the extra cards was present in the former deal. This can be handled by "filling in the blanks" in the rel and aggr structures.

There is also a number of small errors like indexing with an negative value when cards count is 4 and setting upperbound to 13 when there is fewer cards

mortensp:
Closes #114. Now the transposition table can be reused in SolveBoardInternal whenever the new deal has the same trump suit and none of the cards in the new deal was originally in an other hand. When this happens the old deal, that is the deal registered in/corresponding to the transposition table, is updated as the combination of the two deals. In order for this to work i have changed the code so ALL validations is done on dl.remaningcards and postponed the updates of thrd->suits. And two small error was fixed too. 

I wanted to uses this dll to verify each ply in my own DDS algorithms (11 different ones code to compare the differences ), but that would take days to complete before this change - now it can be done in minutes:-) This is also have I have tested my changes in more than 500.000 calls


zzcgumn:
Whether to re-use a transposition table is now handled outside the call to the solver. I think this use case is handled correctly in the 3.0.0 version of the library. Would be much appreciated if you can confirm.

PR #117, opened 2022-01-17
fix build warning int conversion to ULL

ed2k:


zzcgumn:
This should be handled correctly by the 3.0.0 version of the library. At least it builds without the warning.

PR #120, opened 2022-05-11
Fix examples/Makefiles/Makefile_linux

wpeisert:
Compiled examples reports: "error while loading shared libraries: libdds.so: cannot open shared object file: No such file or directory". This PR fixes it.


zzcgumn:
Version 3.0.0 moves away from make files to bazel as the build system. I am happy to consider CMake in addition to bazel but would prefer to stay away from make files.

PR #123, opened 2022-10-04
Update makefiles for MSYS2/MinGW-w64

feiyunw:
Common changes:
- CXX, CXXFLAGS, LDFLAGS, RM, LDFLAGS and LDLIBS are common variables used in GNU make built-in rules.
CXX defaults to 'g++", which will be OK usually.
see GNU Make manual 10.3 Variables Used by Implicit Rules.
- Add -I../include, so the ugly #include statements can be rectified in the future.

src/Makefiles/Makefile_mingw:
- Add common macros for Windows DLLs.
- THREADING & THREAD_LINK: replace boost w/ winapi
	Boost may not be installed, while winapi is essential on MinGW.
- Add MINGW32 support.
- Add -flto -fwhole-program
FYI:
w/o -flto: DLL size 387,584 bytes
w/ -flto (-fuse-linker-plugin): DLL size 409,088 bytes
w/ -flto -fwhole-program: DLL size 296,960 bytes

test/Makefiles/Makefile_mingw:
- Add common macros for Windows exe.
- $(THREAD_LINK): replace boost w/ winapi
- Use .exe suffix for executales.
	MinGW is for building Windows app.
- Add -flto
- Add missing ThreadMgr.cpp to dds_sources.txt

examples/Makefiles/Makefile_mingw:
- Add common macros for Windows exe.
- Add missing .cpp suffix in $(ALL_EXAMPLE_FILES).
- Replace -Wl,--subsystem,windows w/ -mconsole.
- Optimize rules for executables.
- Use .PRECIOUS to preserve .o files.
See GNU Make manual 10.4 Chains of Implicit Rules.

### DIFF
--- a/examples/Makefiles/Makefile_mingw
+++ b/examples/Makefiles/Makefile_mingw
@@ -16,13 +16,7 @@
 
 
 # If your exact compiler name is not given here, change it.
-# CC		= mingw32-g++
-CC		= i686-w64-mingw32-g++
-
-# Use this one to get Windows multi-threading
-# CC_FLAGS	= -O3 -flto -mtune=generic
-# Use this one to get OpenMP multi-threading
-CC_FLAGS	= -O3 -flto -fopenmp -mtune=generic
+# CXX	= g++
 
 # These flags are not turned on by default, but DDS should pass them.
 # Turn them on below.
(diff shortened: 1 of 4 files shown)


zzcgumn:
Version 3.0.0 of the library has support for building linux, windows and macOS although it has mainly been tested on macOS. Do we still need support for mingw, and if so, can you help with a toolchain?

PR #133, opened 2025-08-12
Regression tests for heuristic sorting

zzcgumn:
Defines regression tests for heuristic sorting. 

Also updates a set of rules that are hopefully useful for continue.dev


zzcgumn:
Was meant to push to my local clone.

PR #134, opened 2025-09-15
refactor/speed up heuristic sorting

zzcgumn:
Comparing for per-board performance confirms that the performance reduction for the new heuristic is overhead.\n\nThis PR includes:\n- TimeStatList::Clear() and Scheduler::ClearTiming() to release heavy timing storage before program exit.\n- Runtime per-board timing fallback so  prints per-board timings when DDS_SCHEDULER isn't enabled.\n- Minor BUILD updates for timing library.


zzcgumn:
Opened by mistake, should have been inside my fork.

PR #136, opened 2026-05-17
Publish Version 3 to dds-bridge

zzcgumn:
This pull request does not update the search algorithm at all. It modernises, and modularises, the codebase. The main changes in philosophy are:

1. Memory for transposition tables and other data types are allocated when a double dummy search is started. The caller may optionally retain the table.
2. Modular build structure with per target unit tests.
3. There is a bare-bone Python interface.


sorenhein (commented):
Thanks for this major update

PR #143, opened 2026-05-22
Gtest workaround

tameware:
Fixed failing MacOS build for test targets. All the tests now compile and pass with `bazel build //...` and `bazel test //library/tests/...`


zzcgumn:
I have, finally, managed to add the `-Wno-character-conversion` without breaking the macOS build on CI. It required a bump in the llvm version which in turn forced a bump in the macOS version to get a modern enough standard C++ library.

Thanks for proposing the workaround.

PR #147, opened 2026-05-23
Parallel Solve all Boards and Improved Thread Safety.

zzcgumn:
Parallel processing of a batch of boards had been removed as the code to chunk about static memory in per thread allocations had become quite fragile, and frequently caused crashes on macOS. Introducing a per call SolverContext, with the option for the calling code to supply one for re-use, should make it easier to implement a parallel version.

There were, however, a couple of lingering pieces of statically allocated memory. This did not prevent creating a parallel version of SolveAllBoards, but was dangerous when library functions were called from several threads simultaneously.

Big thanks and kudos to [jdh8](https://github.com/jdh8) for identifying the problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,9 @@ dump.txt
 
 # Generated documentation output
 doxygen_output/
+
+# CCE (code-context-engine)
+# CCE local cache (per-machine, not for version control)
+.cce/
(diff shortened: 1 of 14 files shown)


ThorvaldAagaard:
This is the result of my test
image

PR #157, opened 2026-05-29
Fix AnalysePlay under-counting (issue #156): reuse solver context across cards

### LINKED ISSUE #156: incorrect results with v3
It looks like the refactor broke correctness.
To reproduce, change one of the the hands in `hands.cpp` to:

```
PBN: "S:97652.K632.K8.AQ QJ3.QJ74.QJT93.2 AKT4.5.762.J9864 8.AT98.A54.KT753"
Play: "SKS8S2S3H5HTHKH4H2HQD2H8SQSAHAS5H9H3HJC4H7D6C5H6DQD7DAD8D5DKD3C6S9SJS4C7DJC8D4S6DTC9CTCQD9CJCKS7C2STC3CA"
Play Count: 52
Trump: H
First Hand: N
Dealer Hand: S
Vulnerability: EW
```

Compiling, then running `analyse_play_pbn` gives different results 2.9 vs 3.0.

On v3

```
Number : 49
Play  0: -- 8
Play  1: SK  8
Play  2: S8  7
Play  3: S2  8
Play  4: S3  7
Play  5: H5  8
Play  6: HT  8
Play  7: HK  8
Play  8: H4  8
Play  9: H2  8
Play 10: HQ  8
```

On v2.9

```
Number : 49
Play  0: -- 8
Play  1: SK  8
Play  2: S8  8
Play  3: S2  8
Play  4: S3  8
Play  5: H5  8
Play  6: HT  8
Play  7: HK  9
Play  8: H4  9
Play  9: H2  9
Play 10: HQ  9
```

v2.9 seems to be correct.

ThorvaldAagaard:
## Summary

`AnalysePlay` under-counts tricks for many deals (issue #156). The root cause is in the incremental analyser, not the alpha-beta core — a large `SolveBoard`/`CalcDDtable` differential against 2.9 shows zero differences; only the analyse-play path regressed, which is why `dtest`/`list10000.txt` does not catch it.

`AnalysePlayBin` solves the opening position once, then for each later card calls `analyse_later_board`, which runs a **hint-bounded null-window search**. That search is only correct against the **transposition table built by the initial solve and the preceding cards** — the per-card `hint`/`hintDir` bounds assume a warm TT.

After the context refactor, `analyse_later_board` built a **fresh `SolverContext` per call**:

```cpp
SolverContext ctxLater{thrp};
```

Each `SolverContext` owns its own TT (`SearchContext::tt_`, created lazily), so every card got a **cold, empty TT**. The null-window search then settles on the wrong bound and under-counts. The error is intermittent and "self-heals" (e.g. in #156: wrong at plies 2, 4, 7–15, correct again from 16), because some positions resolve correctly within the bounds regardless of TT state.

This is the same defect already fixed for `calc_dd_table` in 27030ba ("Fix calc DD table consistency by reusing SolverContext across declarers"). `analyse_later_board` is the play-analyser analogue and was missed by that fix.

## Fix

Make `analyse_later_board` take `SolverContext&` and reuse the caller's context (and its warm TT), exactly as `solve_same_board` already does. `AnalysePlayBin` passes its persistent `outer_ctx` to every call. `ctx.search().ini_depth()` still resolves to `thr_->iniDepth` (shared `ThreadData`), so the per-card `--ini_depth` is unchanged; the only behavioral difference is that the TT now persists across the play, as it did before the refactor.

## Test

Adds `library/tests/solve_board/analyse_play_consistency.cpp`: an independent **self-consistency** check that needs no external reference solver — `AnalysePlayPBN`'s trick count after each card must equal a fresh `SolveBoardPBN` of that same position. It covers the exact issue #156 deal plus deterministic random play-outs.

- Fails on the current code (both cases).
- Passes with this fix.
- All existing unit tests still pass.

## Verification

- The #156 deal now matches 2.9 on all 49 plies.
- Random full-play-out differential vs 2.9, 500 deals: current `develop` diverges on **336/500**; with this fix, **0/500**.

Fixes #156


zzcgumn (approved):
Thanks for this. Changes look good to me.

PR #160, opened 2026-05-31
python: bind AnalysePlayPBN, AnalyseAllPlaysPBN, DealerPar, SetMaxThreads

ThorvaldAagaard:
Add Python bindings for four DDS C-API functions that had no dds3 wrapper:

- analyse_play_pbn / analyse_all_plays_pbn -> AnalysePlayPBN / AnalyseAllPlaysPBN, returning {number, tricks} per deal. New solved_play_to_dict converter.
- dealer_par -> DealerPar, returning {score, number, contracts} from a calc_dd_table result.
- set_max_threads -> SetMaxThreads, to size the batch-solver thread pool.

All released the GIL around the C call, validate inputs (play length/parity, seat/vulnerability ranges, board count <= MAXNOOFBOARDS), and reuse the existing pbn_to_deal / dict_to_dd_table_results helpers. Exposed via dds3.__init__, documented in docs/python_interface.md, and covered by python/tests/test_analyse.py (//python:analyse_test). Full suite: 11/11 pass.

### DIFF
--- a/docs/python_interface.md
+++ b/docs/python_interface.md
@@ -286,15 +286,15 @@ Batched form of `analyse_play_pbn` (wraps `AnalyseAllPlaysPBN`).
 
 Legacy thread-resource hook (wraps the **deprecated** `SetMaxThreads` C API).
 
-DDS 3.x removed internal batch parallelism: the legacy batch APIs
-(`solve_all_boards_*`, `analyse_all_plays_pbn`) run **single-threaded** regardless
-of this value, so calling it does **not** enable multi-threaded solving — it only
-sizes the single internal thread's memory and is kept for backward compatibility.
-`user_threads` must be `>= 0` (`0` = auto); values `> 1` are accepted but clamped
-to one thread. Raises `ValueError` for negative values.
+This does **not** control DDS's batch parallelism and is kept only for backward
+compatibility. `solve_all_boards_*` already parallelise across the machine's
+hardware threads automatically (via `solve_boards_n`) — the value passed here does
+not size that pool. `analyse_all_plays_pbn` currently runs sequentially.
+`user_threads` must be `>= 0` (`0` = auto); raises `ValueError` for negative values.
 
-For real concurrency, create one `SolverContext` per worker thread and pass it to
-`solve_board` / `solve_board_pbn` (which release the GIL during the search).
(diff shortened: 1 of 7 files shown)


ThorvaldAagaard:
Good comments, I will update accordingly

docs/python_interface.md
```diff
@@ -228,6 +228,74 @@ print(f"Par: {par_result['par_score']}")
...
+  (str, required), and optional `trump`, `first`, `current_trick_suit`, `current_trick_rank`.
+
+**Returns:**
+- list[dict]: one `{number, tricks}` dict per deal (same shape as `analyse_play_pbn`).
+
+#### `set_max_threads(user_threads=0)`
+
+Legacy thread-resource hook (wraps the **deprecated** `SetMaxThreads` C API).
+
+DDS 3.x removed internal batch parallelism: the legacy batch APIs
```
zzcgumn: `solve_boards_n` has been updated to solve batches in parallel, see line 81 in `solve_board.cpp`.

That said, python applications may have good reasons to handle parallel execution themselves and dds3 makes this possible.

python/src/bindings.cpp
```diff
@@ -594,6 +595,231 @@ auto register_calc_par_bindings(py::module_& module) -> void
...
+            if (user_threads < 0) {
+                throw py::value_error(
+                    "user_threads has invalid value " + std::to_string(user_threads) +
+                    " (expected >= 0; 0 = auto)");
+            }
+            SetMaxThreads(user_threads);
+        },
+        py::arg("user_threads") = 0,
+        "Legacy thread-resource hook (wraps the deprecated SetMaxThreads C API).\n\n"
+        "DDS 3.x removed internal batch parallelism: the legacy batch APIs run\n"
```
zzcgumn: See above.

ThorvaldAagaard:
One open question while we're here: given set_max_threads no longer influences the parallel batch path, would you rather I drop the set_max_threads binding entirely from this PR? It's effectively vestigial now — I only included it for C-API completeness. Easy to remove if you'd prefer not to expose a deprecated knob.

zzcgumn:
Let's merge like this. We should rename the function to something like `InitialiseReadOnlyData` and remove the unused argument. I prefer to this across all builds in once to cause minimum confusion.

PR #163, opened 2026-06-01
First pass at a web interface - includes wasm BUILD refactor

tameware:
See it in action at https://tameware.com/adam/bridge/dds/dds_mvp.html


tameware:
Thanks for looking at this so quickly!

I'm dealing with the Windows WASM errors one by one. It's awkward that I
have to push to github to test. I could start up a PC emulator on my Mac,
but I've been too lazy to do that so far.

This PR is built on top of the previous WASM one. If it's OK with you, I'll
make the fixes here and not cherry-pick them back.

tameware:
Everything builds successfully now, but I'm stlll working on implementing Copilot's suggestions. I tried to turn the PR into a draft, but I was not successful. I'll post another comment when it's ready for human review.

tameware:
Aha! I see I was successful, I just didn't know it. It's a draft - perfect.

tameware:
Someone with Copilot Pro or better, please request a re-review. µSoft won't let me purchase a subscription - they say "New plan sign-ups are temporarily paused as we ensure a high-quality experience."

zzcgumn:
>"New plan sign-ups are temporarily paused as we ensure a high-quality experience."

They gave me a free subscription as an open source contributor a couple of months back. I have found it is very good at code reviews. I guess it has a lot of them to learn from.

zzcgumn:
> If it's OK with you, I'll make the fixes here and not cherry-pick them back.

Perfectly fine with me. Please close the WASM pull request when it is not needed anymore.

tameware:
Copilot issues addressed. Please ask it to have another look.

I added an end-to-end test for the web UI, but it makes bazel test /... slow. Should I look for a way to exclude it from a default test run?

zzcgumn:
There is a mechanism to tag tests and only run those that match the correct tags. In another project I have some tests tagged as `no_ci` and they are excluded from the CI run. I can dig out the details if this is not enough information for Cursor to figure it out.

examples/calc_dd_table_pbn.cpp
```diff
@@ -27,7 +27,7 @@ auto main() -> int
   char line[80];
   bool match;
 
-#if defined(__linux) || defined(__APPLE__)
+#if defined(__linux) || defined(__APPLE__) || defined(__WASM__)
   SetMaxThreads(0);
 #endif
```
tzimnoch, line 32: Out of scope for this PR, but I noticed this in the 2.9.0 version of DDS as well and compensated by calling SetMaxThreads(1) from javascript on the wasm instance after loading it. I can verify whether SetMaxThreads(0) has the same effect.

I feel this is something the library should handle during its initialization. As it stands, failing to call this function before any solve function causes the dds library to crash.

doc/DLL-dds_x.htm has this to say about it currently:
```
The number of threads can be influenced by calling SetMaxThreads.
This function should probably always be called on Linux/Mac, with a zero argument for auto-configuration.
```
and
```
Calling SetMaxThreads is optional, not mandatory.
DDS will always select a suitable number of threads on its own.
```

I have not found this to be true, at least not for wasm binary. It appears to be required to prevent wasm binary from crashing.

web/requirements_lock.txt
```diff
@@ -0,0 +1,105 @@
+#
+# This file is autogenerated by pip-compile with Python 3.14
```
tzimnoch, line 2: I'm not sure what this file is. It looks like it belongs in .gitignore.

tzimnoch (approved):
I think this is all in the right direction, though I'd still move ./examples/wasm to ./wasm. It's just as deserving as the python wrappers.

examples/calc_dd_table_pbn.cpp
tzimnoch: It is possible this was fixed between 2.9.0 and 3.0.0. I have not re-verified, but the inclusion of the call here leads me to believe it's still an issue.

web/requirements_lock.txt
tameware: I wondered the same. After a quick investigation I thought it should stay, but I'll have another look and find a spot to document what I find.

Thanks for the review!

tameware:
[tzimnoch](https://github.com/tzimnoch) wrote:
> I think this is all in the right direction, though I'd still move ./examples/wasm to ./wasm. It's just as deserving as the python wrappers.

I don't disagree. I think the whole layout might deserve a rethink. I'd like to leave this up to Martin. If he wants directories moved I'll be happy to file a PR for it.

examples/calc_dd_table_pbn.cpp
zzcgumn: SetMaxThreads is only initialising some static, read-only memory in 3.0.0. It should be renamed to reflect this. AFAIK all calls to solve a board verifies that this initialisation has been done.

I don't know whether web applications will want to handle threading themselves or call functions like `solve_all_boards_n` which is multithreaded. I don't see any harm in adding an optional `max_threads` parameter to `solve_all_boards_n` if this is desired.

zzcgumn:
>I think this is all in the right direction, though I'd still move ./examples/wasm to ./wasm. It's just as deserving as the python wrappers.

I agree with this but I think it should go in a separate pull request, this one is getting quite big as it is.

tameware:
> [tzimnoch](https://github.com/tzimnoch) wrote:
> 
> > I think this is all in the right direction, though I'd still move ./examples/wasm to ./wasm. It's just as deserving as the python wrappers.

Added as https://github.com/dds-bridge/dds/issues/167

tameware:
> There is a mechanism to tag tests and only run those that match the correct tags. In another project I have some tests tagged as `no_ci` and they are excluded from the CI run. I can dig out the details if this is not enough information for Cursor to figure it out.

Done. Cursor had no trouble with it.

web/requirements_lock.txt
tameware: Cursor confirms that it should stay. I've added extensive comments to web/requirements.in

examples/calc_dd_table_pbn.cpp
tameware: I've moved these comments to Issue https://github.com/dds-bridge/dds/issues/168

tameware:
Copilot should be happy now.

I apologize for the size of this PR! I'm still getting back into the github habit. Per my coding philosophy, the ideal PR is one line, perhaps with comments.

web/dds_mvp.js
```diff
@@ -0,0 +1,336 @@
+// Copyright 2020-2026 Adam Wildavsky
+//
+//   Use of this source code is governed by an MIT-style
+//   license that can be found in the LICENSE file or at
+//   https://opensource.org/licenses/MIT
+
+// TODO: Add tests for the exported functions.
```
zzcgumn, line 7: Has this TODO been addressed? We should either create an issue for the work or remove the comment.

zzcgumn:
I am happy to merge this once the TODO comment has been handled. Tools like Cursor and Claude Code do very easily generate huge pull requests.

tameware:
They're tested indirectly by the e2e test. I'll add an issue for unit tests.

tameware:
> They're tested indirectly by the e2e test. I'll add an issue for JS unit tests.

Actually, not all of the JS was covered by the e2e tests.

Created Issue https://github.com/dds-bridge/dds/issues/169

examples/calc_dd_table_pbn.cpp
tzimnoch: I will verify the extent to which this is still an issue in 3.0.0 and follow up on issue #168 about that.

I'll say here quickly that wasm is inherently single-threaded. You have to set up web workers if you want multithreaded behavior, so I don't expect it to be as simple as passing a parameter to a library call for wasm clients to get multithreaded behavior.

PR #178, opened 2026-06-08
Save build artifacts

tameware:
The e2e tests fail sporadically on github, preventing merges. Adding artifacts will allow us to diagnose and hopefully prevent these failures.


zzcgumn (approved):
Thanks for doing this.

PR #179, opened 2026-06-08
Fix flaky e2e tests

tameware:


tameware:
Please ask Copilot for another review. I had to merge with develop. I'd like to make sure I didn't miss anything. Cursor's agent is happy, but they each catch things the other doesn't.

tameware:
Should be good now.

PR #181, opened 2026-06-12
Fix CalcDdTable error message naming in SolverContext.

tameware:
Use nameof(CalcDdTable) instead of nameof(SolveBoard) so DEBUG exceptions identify the failing method correctly.


zzcgumn (approved):
LGTM

PR #187, opened 2026-06-13
Multithread calc tables

tameware:


library/src/calc_tables.cpp
```diff
@@ -102,13 +106,81 @@ auto calc_all_boards_n(
   return RETURN_NO_FAULT;
 }
 
-// Legacy overload: creates temporary context
+// Legacy overload: parallel across boards, one SolverContext per worker.
 auto calc_all_boards_n(
   Boards * bop,
   SolvedBoards * solvedp) -> int
```
zzcgumn, line 112: We should add an optional argument `user_max_threads`, we can use either std::optional or use a more old school C approach `if (user_max_thread > 0)`.

This has been on my to do list for sometime but I have not come around to get it done.

zzcgumn:
Maybe I am misreading these changes, but I think the calc tables are still single threaded. Looks to me like the parallel execution is solving several boards in parallel.

tameware:
> Maybe I am misreading these changes, but I think the calc tables are still single threaded. Looks to me like the parallel execution is solving several boards in parallel.

You are correct. I know of no way to parallelize the solving of a single deal with a given dealer and vulnerability. Where can I add a comment to make this clearer?

tameware (commented):
Comments addressed.

library/src/calc_tables.cpp
```diff
@@ -102,13 +106,81 @@ auto calc_all_boards_n(
...
+  {
+    SolverContext ctx;
+    return calc_all_boards_n(ctx, bop, solvedp);
+  }
+
+  std::vector<SolverContext> contexts(static_cast<unsigned>(nthreads));
+  std::atomic<int> next_board{0};
+  std::atomic<int> first_error{0};
+
+  auto worker = [&](const int worker_id) {
```
zzcgumn: I think `solve_all_boards_n` and `calc_all_boards_n` only differ by the code in the `worker` lambda and whether passing function pointers or references. Would it be possible to create a `calc_parallel_n` method which accepts three arguments the third being the `worker` lambda?

We can merge these changes as is if Cursor thinks this suggestion is a bad idea.

library/src/system/parallel_boards.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include <algorithm>
+#include <atomic>
+#include <thread>
+#include <vector>
+
+#include <api/dll.h>
+
+
+auto parallel_all_boards_n(
```
zzcgumn, line 20: 👍

zzcgumn (approved):
Thanks for this. I have not had any use cases for solving batches of boards but I know other downstream projects do.

PR #192, opened 2026-06-16
Add Link-Time Optimization for MacOS builds.

tameware:
Speeds up calc runtime by a few percent, with a negligible slowdown in full builds. Would likely help on Windows and Linux as well - someone with those environments could test that.

WASM builds already include full lto.


zzcgumn (approved):
Fair enough.

PR #195, opened 2026-06-17
Fixes so that DDS_AB_STATS will work when it's set.

tameware:
I've tested this locally with `bazel build //library/tests:dtest --cxxopt=-DDDS_AB_STATS`. A dtest run then produces ABstats*.txt files, as it ought to. The build fails without this PR.

We should either merge this PR or else remove the DDS_AB_STATS code entirely.

Another possible enhancement would be to mirror the DDS_SCHEDULER pattern, so that we can use `--define=ab_stats=true` instead of `--cxxopt`.

### DIFF
--- a/library/src/init.cpp
+++ b/library/src/init.cpp
@@ -163,14 +163,6 @@ void InitDebugFiles()
 }
 
 
-void CloseDebugFiles()
-{
-  // Per-context debug files are opened when SolverContext owns its ThreadData
-  // (SolverContext(SolverConfig)) and closed in ~SolverContext() when the last
-  // shared_ptr to that ThreadData is released.
-}
-
-
 void SetDeal(
   const std::shared_ptr<ThreadData>& thrp)
 {
--- a/library/src/init.cpp
+++ b/library/src/init.cpp
@@ -166,7 +166,8 @@ void InitDebugFiles()
 void CloseDebugFiles()
 {
   // Per-context debug files are opened when SolverContext owns its ThreadData
-  // (SolverContext(SolverConfig)) and closed in ~SolverContext() for owners only.
+  // (SolverContext(SolverConfig)) and closed in ~SolverContext() when the last
+  // shared_ptr to that ThreadData is released.
 }
 
 
--- a/library/src/init.cpp
+++ b/library/src/init.cpp
@@ -165,12 +165,8 @@ void InitDebugFiles()
 
 void CloseDebugFiles()
 {
-  for (unsigned thrId = 0; thrId < memory.NumThreads(); thrId++)
-  {
-  SolverContext tmp_ctx;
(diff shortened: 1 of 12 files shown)


zzcgumn:
I think we should prefer `--define=ab_stats=true` over `--cxxopt=-DDDS_AB_STATS`. Explicit is better than implicit and the `--define` pattern is immediately obvious but the `--cxxopt=` forces me parse through the right hand side to understand what is happening.

zzcgumn:
I will give this a more careful review once I have a bit more time. We are out of Copilot cycles for the month of June.

tameware:
There's no hurry!

In any case, I suggest the `--define=ab_stats` refactor can go in a separate PR. It's orthogonal to this change.

library/src/system/file.hpp
```diff
@@ -0,0 +1,45 @@
...
+    void SetName(const std::string& fname_in);
+
+    std::ofstream& GetStream();
+
+    void Close();
+};
+
+}  // namespace dds
+
+using dds::File;
```
zzcgumn: I think this should be in the file.cpp file.

library/src/init.cpp
```diff
@@ -165,12 +165,8 @@ void InitDebugFiles()
 
 void CloseDebugFiles()
```
zzcgumn: Can we remove this empty method?

tameware: Done.

library/src/system/file.hpp
tameware: Removed entirely and qualified all uses, per Cursor's suggestion.

library/src/solver_context/solver_context.cpp
```diff
@@ -1,13 +1,43 @@
...
+  // Ensure persistent facades like SearchContext see the bound ThreadData.
+  search_.set_thread(thr_);
+  search_.set_owner(this);
+  if (!thr_) return;
+
+#if defined(DDS_TOP_LEVEL) || defined(DDS_AB_STATS) || defined(DDS_AB_HITS) || \
+    defined(DDS_TT_STATS) || defined(DDS_TIMING) || defined(DDS_MOVES)
+  thr_->init_debug_files(next_debug_file_suffix());
+#endif
+}
```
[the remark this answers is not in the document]
tameware: Addressed differently, per Cursor.

library/src/system/file.cpp
```diff
@@ -0,0 +1,46 @@
...
+std::ofstream& File::GetStream()
+{
+  if (!file_open_)
+  {
+    fout_.open(fname_);
+    file_open_ = true;
+  }
+
+  return fout_;
+}
```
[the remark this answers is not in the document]
tameware: Addressed differently, per Cursor.

zzcgumn (approved):
Looks good to me.

PR #196, opened 2026-06-18
ASan TSan

tameware:
Fix `ASan` (AddressSanitizer) and `TSan` (ThreadSanitizer) builds on macOS. These are used for testing.

### DIFF
--- a/library/src/system/scheduler.cpp
+++ b/library/src/system/scheduler.cpp
@@ -450,24 +450,24 @@ bool Scheduler::SameHand(
 // that they scale somewhat proportionally to other cases.
 // The strength parameter is currently not used.
 
-int SORT_SOLVE_TIMES[2][8] =
+static int SORT_SOLVE_TIMES[2][8] =
 {
   { 284000,  91000, 37000, 23000, 17000, 15000, 13000, 4000 },
   { 388000, 140000, 60000, 40000, 30000, 23000, 18000, 6000 },
 };
 
-#define SORT_SOLVE_STRENGTH_CUTOFF 0
-
-double SORT_SOLVE_STRENGTH[2][3] =
-{
-  { 1.525, 1.810, 0.0285 },
-  { 1.585, 1.940, 0.0354 }
-};
+// #define SORT_SOLVE_STRENGTH_CUTOFF 0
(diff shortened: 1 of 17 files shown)


zzcgumn:
This looks useful. Xcode's llvm version was far too old when I first tried to add the ASAN and TSAN options.

tameware:
This PR includes several sets of commits that could be submitted separately:

1. Fix the broken ASan and TSan configurations for macOS. They can work now that Xcode's llvm is more up to date.
2. Add ASan and TSan builds to CI.
3. Set all CI timeouts to 20 minutes. Most had the default of six hours.
4. Fix the unsafe code that ASan identified.

It was expedient to do this together. Let me know if you'd rather see separate PRs for each.

library/src/system/scheduler.cpp
```diff
@@ -450,24 +450,24 @@ bool Scheduler::SameHand(
...
 };
 
-#define SORT_SOLVE_STRENGTH_CUTOFF 0
-
-double SORT_SOLVE_STRENGTH[2][3] =
-{
-  { 1.525, 1.810, 0.0285 },
-  { 1.585, 1.940, 0.0354 }
-};
+// #define SORT_SOLVE_STRENGTH_CUTOFF 0
```
zzcgumn, line 459: We should either the deleted the commented out code or explain why we are keeping it.

zzcgumn:
This looks good in general, one small comment from me. I noticed that the building with the ASAN/TSAN flags picked up a couple issues.

library/src/system/scheduler.cpp
tameware: Agreed. That seems orthogonal to this PR. Shall I add an issue for it?

tameware: Also, do you have any reason to keep it?

zzcgumn: No reason to keep it. We can create a new issue to keep this PR leaner.

tameware:
Perhaps I’m confused. Isn’t the PR leaner if we leave this code as is? I’m
away from my laptop, so not seeing the full context.

PR #201, opened 2026-06-19
Fix DDS_SCHEDULER build consistency and dtest -r reporting.

tameware:
Propagate scheduler define through :dds to fix Scheduler ODR/crash; use SolveAllBoards for parallel solve timing; clarify calc has no per-board data.

Test with

`bazel build --define=scheduler=true //library/tests:dtest`
`bazel-bin/library/tests/dtest -f hands/list10.txt -r`


zzcgumn:
Looks good to me but we I think we may have some overlap between this pull request and #204. I am happy to merge this first and resolve potential conflicts in my branch. Does this make sense?

zzcgumn (approved):

tameware:
I'm fine either way. Cursor made these changes - it can do the same again if the merge seems difficult.

PR #204, opened 2026-06-21
Rename SetMaxThreads to InitialiseStaticMemory; add max_threads override

### LINKED ISSUE #168: Confirm whether SetMaxThreads() call is necessary
Per [tzimnoch](https://github.com/tzimnoch)
[examples/calc_dd_table_pbn.cpp](https://github.com/dds-bridge/dds/pull/163/files/08342f1380daa2c2b95cb69c2808f3383ec14f21#diff-601a884b3c1e950973180fd4d4d38bd24a133a9435ed02c610ce3c620c68c49d)
Comment on lines +30 to 32
#if defined(__linux) || defined(__APPLE__) || defined(__WASM__)
  SetMaxThreads(0);
#endif

(https://github.com/dds-bridge/dds/pull/163#discussion_r3352064928)
Out of scope for this PR, but I noticed this in the 2.9.0 version of DDS as well and compensated by calling SetMaxThreads(1) from javascript on the wasm instance after loading it. I can verify whether SetMaxThreads(0) has the same effect.

I feel this is something the library should handle during its initialization. As it stands, failing to call this function before any solve function causes the dds library to crash.

doc/DLL-dds_x.htm has this to say about it currently:

The number of threads can be influenced by calling SetMaxThreads.
This function should probably always be called on Linux/Mac, with a zero argument for auto-configuration.
and

Calling SetMaxThreads is optional, not mandatory.
DDS will always select a suitable number of threads on its own.
I have not found this to be true, at least not for wasm binary. It appears to be required to prevent wasm binary from crashing.

(https://github.com/dds-bridge/dds/pull/163#discussion_r3352157052)
• 
It is possible this was fixed between 2.9.0 and 3.0.0. I have not re-verified, but the inclusion of the call here leads me to believe it's still an issue.

### LINKED ISSUE #188: Add an optional user_max_threads argument to calc_all_boards_n
Per @zzcgumn:

_we can use either std::optional or use a more old school C approach if (user_max_thread > 0)._

zzcgumn:
## Summary
- Rename `SetMaxThreads` to `InitialiseStaticMemory`: the function initialises
  static library state (TT memory pools, scheduler/thread-manager, lookup
  tables) and no longer controls thread count, since internal batch threading
  was removed. `SetMaxThreads` is kept as a thin deprecated alias so the
  exported C symbol, the .NET P/Invoke, and the Python binding keep working.
- Collapse the duplicated `max(1, min(hardware_concurrency, n))` worker-count
  formula into one `resolve_worker_count(max_threads, count)` helper, and plumb
  an optional `max_threads` (0 = auto) through `solve_all_boards_n` and the
  legacy parallel `calc_all_boards_n`.
- Add public `*N` C variants — `SolveAllBoardsN`, `SolveAllBoardsBinN`,
  `CalcDDtableN`, `CalcDDtablePBNN`, `CalcAllTablesN`, `CalcAllTablesPBNN` —
  that accept `maxThreads`. The existing functions keep their signatures/ABI
  and delegate with `0`, so current auto-sizing behaviour is unchanged.
- Add tests: `resolve_worker_count` edge cases, and equivalence checks proving
  an explicit `maxThreads` cap produces results identical to the auto path.

## Notes
- The C++ `calc_dd_table` overloads use the *sequential* context-aware
  `calc_all_boards_n`, where a thread cap would be a no-op, so they were left
  unchanged. Only the legacy parallel path (used by the C API) carries the
  parameter.
- No behaviour change for any existing caller: defaults and delegation preserve
  the current `hardware_concurrency` auto-sizing.

## Test plan
- [x] `bazel build //library/... //examples/...`
- [x] `bazel test //library/tests/...` — 31 tests pass (29 existing + 2 new)
- [x] `CalcDDtableN(..., maxThreads=1)` matches `CalcDDtable(...)`
- [x] `SolveAllBoardsBinN(..., maxThreads=1)` matches `SolveAllBoardsBin(...)`
- [x] `InitialiseStaticMemory()` and the deprecated `SetMaxThreads` alias both
      leave the library usable for a subsequent solve

Closes #168
Closes #188

### DIFF
--- a/examples/dd_table_for_deal.cpp
+++ b/examples/dd_table_for_deal.cpp
@@ -227,7 +227,7 @@ auto main(int argc, char * argv[]) -> int
   char line[80];
 
 #if defined(__linux) || defined(__APPLE__)
-  SetMaxThreads(0);
+  InitialiseStaticMemory();
 #endif
 
   const int res = CalcDDtablePBN(tableDealPBN, &table);
--- a/examples/dd_table_for_deal.cpp
+++ b/examples/dd_table_for_deal.cpp
@@ -226,10 +226,6 @@ auto main(int argc, char * argv[]) -> int
   DdTableResults table;
   char line[80];
 
-#if defined(__linux) || defined(__APPLE__)
-  InitialiseStaticMemory();
-#endif
-
   const int res = CalcDDtablePBN(tableDealPBN, &table);
   if (res != RETURN_NO_FAULT)
   {
--- a/dotnet/DDS_Core/DDS.cs
+++ b/dotnet/DDS_Core/DDS.cs
@@ -31,7 +31,7 @@ public class DDS
         /// </remarks>
         /// <param name="userThreads">Ignored; retained for backward compatibility.</param>
         // The underlying C symbol SetMaxThreads was renamed to
-        // InitialiseStaticMemory; this P/Invoke still targets the deprecated alias.
+        // InitializeStaticMemory; this P/Invoke still targets the deprecated alias.
         [Obsolete("Use SolverContext instead.")]
         public void SetMaxThreads(int userThreads) => DdsNative.SetMaxThreads(userThreads);
 
--- a/dotnet/DDS_Core/DDS.cs
+++ b/dotnet/DDS_Core/DDS.cs
@@ -29,7 +29,7 @@ public class DDS
         /// via <c>SolverContext</c> and <c>SolverConfig</c>.
         /// </para>
         /// </remarks>
-        /// <param name="userThreads">Maximum number of threads to use.</param>
+        /// <param name="userThreads">Ignored; retained for backward compatibility.</param>
         // The underlying C symbol SetMaxThreads was renamed to
         // InitialiseStaticMemory; this P/Invoke still targets the deprecated alias.
         [Obsolete("Use SolverContext instead.")]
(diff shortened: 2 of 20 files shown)


tameware:
Shall I wait to request another Copilot review until the build is fixed?

zzcgumn:
Ready for another round of Copilot.

tameware (commented):
I am not qualified to review the code for readability and functionality - I'm happy to delegate that to Copilot.

I did pull the branch and run it through my benchmarks. I see that dtest's `-n` flag - number of threads - still works, so the replacement threading limits are being respected.

Copilot is complaining about something that a previous iteration marked as resolved. I suggest either making Copilot happy or adding a comment noting why it's OK.

The spelling of Initialise is inconsistent with the internal variable spelling that uses the American Initialize. I suggest renaming to InitializeStaticMemory everywhere unless there's a reason not to. That's what most developers will expect.

examples/dd_table_for_deal.cpp
```diff
@@ -227,7 +227,7 @@ auto main(int argc, char * argv[]) -> int
   char line[80];
 
 #if defined(__linux) || defined(__APPLE__)
-  SetMaxThreads(0);
+  InitialiseStaticMemory();
```
tameware: I think this call can be removed. No matter which platform, it's already happened.

tameware:
LGTM

zzcgumn:
@tameware, would you mind approving the pull request as well?

tameware:
I would if I knew how! I see no "Approve" button.

zzcgumn:
Do you see a submit review button here https://github.com/dds-bridge/dds/pull/204/changes?

Otherwise I need to look into permissions.

tameware (approved):

PR #211, opened 2026-06-27
Remove unused itest.cpp test driver.

tameware:
The file was excluded from the dtest build, had no Bazel target, and referenced a nonexistent realMain entry point.


zzcgumn (approved):
Thanks for cleaning this up.

PR #215, opened 2026-06-28
Deduplicate identical deals in CalcAllTablesN

tameware:
CalcAllTablesN expanded every deal x strain into a board and solved them all, even when a batch contained identical deals. v2.9 deduplicates identical deals (crossref/CopyCalcSingle), solving each distinct deal once and copying the table to the copies.

Detect duplicate deals by card content, solve only the unique deals, and map every deal (unique or copy) back to its canonical solve when filling the result tables. Identical deals always yield identical double-dummy tables, so this is exact.

On hands/list1000.txt (36% duplicate deals) single-threaded calc drops from ~104s to ~68s of user time, matching v2.9 (~65s); the residual gap versus v2.9 is parallel scaling, addressed separately. All library tests pass and dtest's table validation is unchanged.

### DIFF
--- a/library/src/calc_tables.cpp
+++ b/library/src/calc_tables.cpp
@@ -220,6 +220,18 @@ int STDCALL CalcDDtable(
 }
 
 
+namespace
+{
+auto same_deal_cards(const DdTableDeal& a, const DdTableDeal& b) -> bool
+{
+  for (int h = 0; h < DDS_HANDS; h++)
+    for (int s = 0; s < DDS_SUITS; s++)
+      if (a.cards[h][s] != b.cards[h][s])
+        return false;
+  return true;
+}
+}
+
 int STDCALL CalcAllTablesN(
   DdTableDeals const * dealsp,
   int mode,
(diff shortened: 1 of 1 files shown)


library/src/calc_tables.cpp
```diff
@@ -254,12 +266,40 @@ int STDCALL CalcAllTablesN(
...
+    }
+    if (found < 0)
+    {
+      deal_to_unique[static_cast<unsigned>(m)] =
+        static_cast<int>(unique_deals.size());
+      unique_deals.push_back(m);
+    }
+    else
+      deal_to_unique[static_cast<unsigned>(m)] = found;
+  }
```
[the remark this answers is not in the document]
tameware: Initial benchmarking shows this change may do more harm than good. On the back-burner for now.

library/src/calc_tables.cpp
```diff
@@ -220,6 +220,18 @@ int STDCALL CalcDDtable(
 }
 
 
+namespace
+{
+auto same_deal_cards(const DdTableDeal& a, const DdTableDeal& b) -> bool
```
zzcgumn, line 225: I would define this as `operator==(const DdTableDeal& a, const DdTableDeal& b) -> bool` even though DdTableDeal is a C struct. This file is after all pure C++.

Would Cursor disagree?

tameware: I'll ask if you like, but we are abandoning this PR as, per our email discussion, we don't expect identical deals in practice.

zzcgumn: Fair enough :)

zzcgumn:
@tameware, Could you please add a comment that we don't expect this to be a problem for real world data sets and close this pull request?

tameware:
We don't expect this to be a problem for real world data sets.

PR #216, opened 2026-06-29
Improve parallelism by solving most difficult deals first

tameware:
```
$ time ./benchmark.sh --branch opus-two-percent --branch opus-improve-parallelism --repeats 5 --max-deals 10000
Building dtest from 'opus-two-percent'...
Building dtest from 'opus-improve-parallelism'...
Restoring 'benchmark'...
DDS dtest benchmark
===================
branch:      branch 'opus-two-percent' (/var/folders/12/xtx6dlwd0mdcxkspvmsszsrc0000gn/T//dds-dtest-branch.Cf0Cly)
compare:     branch 'opus-improve-parallelism' (/var/folders/12/xtx6dlwd0mdcxkspvmsszsrc0000gn/T//dds-dtest-compare.gwQNLJ)
details:     off (summary only)
run order:   interleaved branch, compare
epsilon:     0.5%
hands dir:   /Users/adamw/src/dds/hands
max_deals:   10000
files:       list10000.txt list1000.txt list100.txt list10.txt list1.txt
git branch:  benchmark
repeats:     5


Summary (avg user ms)
==============================================================================
solver file          opus-improve opus-two-per cmp/branch note
------ ------------- ------------ ------------ ---------- ---------------
solve  list10000.txt         2.67         2.63      1.01x opus-two-percent faster
solve  list1000.txt          2.12         2.09      1.01x opus-two-percent faster
solve  list100.txt           2.06         2.05      1.00x equal
solve  list10.txt            4.20         4.30      0.98x opus-improve-parallelism faster
solve  list1.txt            12.00        12.00      1.00x equal
calc   list10000.txt         9.68        10.25      0.94x opus-improve-parallelism faster
calc   list1000.txt          9.63        10.64      0.91x opus-improve-parallelism faster
calc   list100.txt           8.22         8.70      0.95x opus-improve-parallelism faster
calc   list10.txt           14.44        14.38      1.00x equal
calc   list1.txt            38.20        38.00      1.01x opus-two-percent faster
------ ------------- ------------ ------------ ---------- ---------------
TOTAL  elapsed (s)         684.60       716.16      0.96x opus-improve-parallelism faster

Completed 100 runs (100 expected).
```


library/src/system/parallel_boards.cpp
```diff
@@ -31,23 +31,51 @@ auto resolve_worker_count(
 }
 
 
+static auto is_permutation_of_range(
+  const std::vector<int>& order,
+  const int count) -> bool
+{
```
zzcgumn, line 37: We have `std::set_intersection` and `std::set_difference` available in <algorithm>. You can use `std::ranges::views::iota` to generate a range from `[0..count]`.

Standard library functions are at least expected to be faster and reduce the amount of code we need to write.

tameware: Nice find! Fixed.

tameware: Or not…

Copilot reports that the change is less efficient:

is_permutation_of_range() currently re-sorts the order vector and builds an intermediate common vector via set_intersection. This is O(n log n) plus extra allocations and will run on every parallel_all_boards_n() invocation that supplies an order (e.g. batch calc), adding overhead in a performance-sensitive path. You can validate a permutation of [0, count) in O(n) without sorting by tracking seen indices and checking bounds/duplicates.

Thoughts, @zzcgumn?

library/src/calc_tables.cpp
```diff
@@ -8,26 +8,28 @@
...
 #include <solver_if.hpp>
+#include <lookup_tables/lookup_tables.hpp>
 #include <system/memory.hpp>
 #include <system/parallel_boards.hpp>
 #include <system/scheduler.hpp>
 #include <system/system.hpp>
 
 
 extern Memory memory;
-extern Scheduler scheduler;
```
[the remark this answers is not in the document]
tameware: Fixed without Copilot - I don't like the way it creates a new branch.

library/src/system/parallel_boards.cpp
tameware: I've had another look. I agree with Copilot - I prefer the original implementation:

```
static auto is_permutation_of_range(
  const std::vector<int>& order,
  const int count) -> bool
{
  std::vector<char> seen(static_cast<unsigned>(count), 0);
  for (const int v : order)
  {
    if (v < 0 || v >= count || seen[static_cast<unsigned>(v)])
      return false;
    seen[static_cast<unsigned>(v)] = 1;
  }
  return true;
}
```

`seen` could equally well be `std::vector<unsigned char> `

library/src/calc_tables.cpp
```diff
@@ -8,12 +8,15 @@
...
 #include "calc_tables.hpp"
+#include <algorithm>
+#include <numeric>
 #include <vector>
 
 #include <pbn.hpp>
 #include <solve_board.hpp>
 #include <api/solve_board.hpp>
 #include <solver_if.hpp>
+#include <lookup_tables/lookup_tables.hpp>
```
tzimnoch: I should probably pull and compile, but I don't see what's being used from this header. fanout comes from scheduler.hpp and all the other called functions appear to be in std:: namespace. What did I miss?

library/src/calc_tables.cpp
```diff
@@ -192,7 +215,8 @@ int STDCALL CalcDDtableN(
     ind++;
   }
 
-  int res = calc_all_boards_n(&bo, &solved, maxThreads);
+  // Single deal: all boards share one deal, so hardest-first sorting is a no-op.
+  int res = calc_all_boards_n(&bo, &solved, maxThreads, /*difficulty_sort=*/false);
```
tzimnoch, line 220: trivial

Inline variable-name comment isn't necessary.

library/src/system/scheduler.cpp
```diff
@@ -758,6 +758,12 @@ int Scheduler::Strength(const Deal& dl) const
 
 
 int Scheduler::Fanout(const Deal& dl) const
+{
+  return deal_fanout(dl);
+}
+
+
+auto deal_fanout(const Deal& dl) -> int
```
tzimnoch: I don't have any skin in this, but why not move Fanout from private to public instead?

What are the advantages of exposing the function this way? (Ideological purity being a perfectly fine answer.)

library/src/system/parallel_boards.cpp
tzimnoch: set_difference makes sense if you need to know what the difference is. The use case here doesn't need to prove as hard a problem.

library/src/ab_search.cpp
```diff
@@ -878,10 +878,10 @@ void make_3(
...
-  posPoint->winner[st].rank = static_cast<unsigned char>(thrp->rel[aggr].abs_rank[1][st].rank);
-  posPoint->winner[st].hand = static_cast<unsigned char>(thrp->rel[aggr].abs_rank[1][st].hand);
-  posPoint->second_best[st].rank = static_cast<unsigned char>(thrp->rel[aggr].abs_rank[2][st].rank);
-  posPoint->second_best[st].hand = static_cast<unsigned char>(thrp->rel[aggr].abs_rank[2][st].hand);
+  posPoint->winner[st].rank = thrp->rel[aggr].abs_rank[1][st].rank;
+  posPoint->winner[st].hand = thrp->rel[aggr].abs_rank[1][st].hand;
+  posPoint->second_best[st].rank = thrp->rel[aggr].abs_rank[2][st].rank;
+  posPoint->second_best[st].hand = thrp->rel[aggr].abs_rank[2][st].hand;
 
     }
```
tzimnoch: Might as well fix the indenting problem at the same time.

tameware:
When reading doc/2014-11 Performance and Benchmarking.docx I learned that the Scheduler should be doing this already. I asked Cursor about it. It replied:

The 2014 doc is right about what the **Scheduler was designed to do**, but at the time of `b08925f` that design was **not actually driving parallel dispatch** on the calc path. The commit worked because it reintroduced hardest-first ordering in the only place that mattered for batch calc.

## What the commit changed

Before `b08925f`, parallel calc used `parallel_all_boards_n`, which hands out work with an atomic counter in **board index order** (0, 1, 2, …):

```101:106:library/src/system/parallel_boards.cpp
      const int slot = next.fetch_add(1, std::memory_order_relaxed);
      if (slot >= count || first_error.load(std::memory_order_relaxed) != RETURN_NO_FAULT)
      {
        break;
      }
      const int bno = board_of(slot);
```

`b08925f` added an optional permutation so slot 0, 1, 2… map to board numbers sorted by descending `deal_fanout` (same structural proxy as `Scheduler::Fanout`).

## Why that improved wall time (~13%) without changing CPU

This is classic **longest-processing-time-first (LPT)** scheduling for a thread pool.

The 2014 benchmarking doc describes exactly this failure mode: solve times are highly skewed (roughly Weibull), so with work-stealing in index order one thread can draw a very hard board near the end and sit alone while the others idle. Wall time is dominated by that **tail**, not by total CPU.

The commit’s benchmark confirms that story:

- Wall time: ~11.0s → ~9.6s (~13%)
- User CPU: unchanged

So no less work was done — boards were just started in a better order so threads finish closer together.

## “But the Scheduler already sorts hardest first…”

Yes — **in the v2.8 design**, documented in the 2014 paper. The Scheduler:

1. Groups boards (strain + hash, with repeat detection)
2. Predicts difficulty per group via fanout (`SortCalc` / `SortSolve`)
3. Dispatches through `GetNumber()`, which hands out hardest groups first

That old path looked like:

```cpp
st = scheduler.GetNumber(thr_id);
index = st.number;
```

But on the main line at `b08925f`:

| Path | Scheduler sort computed? | Used for parallel dispatch? |
|------|--------------------------|-----------------------------|
| **calc** (`CalcAllTables`) | No — `calc_tables.cpp` never called `RegisterRun` or `GetNumber` | No — index order only |
| **solve** (`SolveAllBoards`) | Yes — `RegisterRun` still runs | No — `GetNumber` was removed when parallelization moved to `parallel_all_boards_n` (May 2026) |

`GetNumber()` still exists in `scheduler.cpp` but is **not called anywhere** on the main branch. The Scheduler’s sort is effectively dead code for dispatch today.

There was a side branch (`a71b01a`, “Restore scheduler-based calc threading…”) that wired calc back through `GetNumber`, but it was **not merged** into the line that contains `b08925f`.

So the doc describes the **intended** v2.8 behavior; `b08925f` fixes a **regression gap** where parallel calc had been running in naive index order.

## How `b08925f` differs from the full Scheduler

The commit is a **lighter-weight** version of the same idea:

- **Per-board** fanout sort, not per-group scheduling with strain/hash buckets
- **No repeat-hand dedup** via scheduler groups (calc comment says repeats are always a direct copy anyway)
- **Trump-independent** fanout only — reasonable for batch calc where many distinct deals dominate
- Explicitly **skipped for single-deal** `CalcDDtable` (all boards share one deal → same fanout → sort is a no-op)

The full Scheduler’s `SortCalc` uses the same fanout concept but with NT/suit-specific scaling curves; the commit uses the raw fanout score, which is enough to fix the tail problem.

## Bottom line

`b08925f` helped because parallel calc had lost hardest-first dispatch somewhere between the v2.8 Scheduler design and the newer `parallel_all_boards_n` helper. The 2014 doc is describing behavior that **used to be wired up** via `GetNumber`; `b08925f` restores the essential part of that benefit (hard boards first) for calc without reintegrating the full Scheduler machinery.

tameware:
As far as I can tell, a71b01a, “Restore scheduler-based calc threading…”, was never merged and is not part of a current PR. We should merge either this PR or one containing a71b01a, not both.

tameware:
I see now that https://github.com/dds-bridge/dds/commit/a71b01a0e308ccf558a698a2d5bbd4c5cc2c04f3 was mine! I think I abandoned it because I mistakenly thought it degraded performance. That was before I'd vibe-coded benchmark.sh. 

I cherrypicked https://github.com/dds-bridge/dds/commit/a71b01a0e308ccf558a698a2d5bbd4c5cc2c04f3 on top of the develop head and ran benchmark.sh, averaging over three runs. It shows timing close between the two approaches:

```
Summary (avg user ms)
==============================================================================
solver file               develop calc-run-thr opus-improve
------ ------------- ------------ ------------ ------------
solve  list1000.txt          1.54         1.56         1.55
solve  list100.txt           1.99         1.94         1.93
solve  list10.txt            4.27         4.17         4.30
solve  list1.txt            12.67        13.33        12.00
calc   list1000.txt         11.21         9.68        10.07
calc   list100.txt           9.24         8.23         8.67
calc   list10.txt           16.10        14.97        14.90
calc   list1.txt            41.67        41.00        39.33
------ ------------- ------------ ------------ ------------
TOTAL  elapsed (s)          42.89        38.14        39.29
```

tameware:
Thanks for the review, @tzimnoch!

I'll hold off on these changes until we hear from @zzcgumn as to which approach he'd prefer, this one or _Restore scheduler-based calc threading_.

zzcgumn (approved):
I think we go with this pull request for sorting batches to start with the most difficult hand to solve.

tameware:
OK! I'll make the changes suggested by [tzimnoch](https://github.com/tzimnoch), including removing the less-efficient STL use.

library/src/ab_search.cpp
tameware: Addressed.

library/src/system/scheduler.cpp
tameware: Nice catch. Here's what Cursor has to say:

> deal_fanout depends only on Deal, not on Scheduler state. Keeping it free:
> 
> avoids requiring a Scheduler instance,
> makes the stateless nature explicit,
> allows reuse/testing independently of scheduler lifecycle.
> However, the forwarding Scheduler::Fanout is redundant. I’d remove it and keep deal_fanout as an internal utility, ideally in a more appropriate utility header/source. Making Fanout public would unnecessarily suggest it is scheduler behavior.

I'll figure out a spot for the function, maybe as a method of Deal.

library/src/calc_tables.cpp
tameware: Why not? It seems useful to me. Were I writing the code I might have replaced it with a named constant, say DIFFICULTY_SORT_FALSE.

library/src/system/scheduler.cpp
tameware: Done. Added deal.cpp, since this is the first method on Deal.

library/src/calc_tables.cpp
tameware: It's now in deal.cpp. Per Cursor:

> Deal::fanout() uses:
> 
> group_data[c].last_group_
> group_data and its MoveGroupType element type are declared in lookup_tables.hpp. The direct include is therefore required and avoids relying on a transitive include.

Should I add a comment above the #include?

library/src/system/parallel_boards.cpp
```diff
@@ -31,23 +33,54 @@ auto resolve_worker_count(
...
+
+  const auto expected = std::views::iota(0, count);
+
+  std::vector<int> common;
+  common.reserve(static_cast<unsigned>(count));
+  std::ranges::set_intersection(
+    sorted, expected, std::back_inserter(common));
+
+  return static_cast<int>(common.size()) == count;
+}
```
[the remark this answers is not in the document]
tameware: Reverted to previous, more efficient implementation.

library/src/system/parallel_boards.hpp
```diff
@@ -28,9 +29,15 @@ auto resolve_worker_count(int max_threads, int count) -> int;
  * @param worker_cap Maximum worker threads; <= 0 uses hardware concurrency.
  * @param process_board Called for each board; must return RETURN_NO_FAULT (1)
  *        on success. Receives the worker's thread index and board number.
+ * @param order Optional dispatch order: a permutation of [0, count) giving the
+ *        sequence in which board numbers are handed out (e.g. hardest first to
+ *        shorten the tail). When null/empty, boards are dispatched in index
+ *        order. Only the dispatch order changes; @p process_board still receives
+ *        the real board number, so result placement is unaffected.
  * @return First non-success code from @p process_board, or RETURN_NO_FAULT.
```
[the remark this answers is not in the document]
tameware: Addressed.

tameware:
Latest benchmark with this PR: Calc for 1,000 unique deals (so 20,000 solves) is 9% faster than the develop head, 9% slower than v 2.9.

library/src/calc_tables.cpp
tzimnoch: No, if it's necessary it's necessary. I missed what new function was being used.

library/src/api/dll.h
```diff
@@ -186,6 +186,16 @@ struct Deal
...
   unsigned int remainCards[DDS_HANDS][DDS_SUITS];
+
+#ifdef __cplusplus
+  /**
+   * @brief Cheap structural difficulty estimate (cards only, trump-independent).
+   *
+   * Per hand, sum the number of card groups per suit, with a bonus for voids.
+   * Used to dispatch the hardest boards first in parallel batch calc.
+   */
+  auto fanout() const -> int;
```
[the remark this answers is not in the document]
tameware: Would not have changed the ABI, but proper to address anyway. Done.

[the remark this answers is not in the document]
tameware: Made it internal. Darn - I prefer the OO style. But compatibility is important.

library/tests/system/parallel_boards_test.cpp
```diff
@@ -0,0 +1,123 @@
...
+}  // namespace allocation_tracking
+
+void* operator new(const std::size_t size)
+{
+  if (allocation_tracking::enabled.load(std::memory_order_relaxed))
+    allocation_tracking::allocations.fetch_add(1, std::memory_order_relaxed);
+
+  if (void* const memory = std::malloc(size))
+    return memory;
+  throw std::bad_alloc();
```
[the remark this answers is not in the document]
tameware: Addressed, though per Cursor not so important:

> The overridden global operator new calls std::malloc(size) directly; when size == 0, this can return nullptr and you’ll throw std::bad_alloc(), but the C++ global operator new is required to succeed for zero-size allocations (treating it as at least 1 byte). This can cause flaky test failures if any code path allocates 0 bytes.
> 
> Technically right, practically low priority for this test.
> 
> The standard does require a successful zero-size operator new to return non-null, and malloc(0) is allowed to return nullptr, so this replacement is not fully conforming. Copilot’s “flaky” framing is a bit strong: on platforms where malloc(0) returns null, a zero-size new would throw bad_alloc and fail hard, not flake on the allocation-count assertion.
> 
> In this test binary the tracked path always allocates vector<char>(count) with count == 4, so that path won’t request zero bytes. The bigger caveat is that this operator new is global for the whole test executable, so some other zero-size allocation (gtest/library internals) could trip it on a strict CRT.

library/src/calc_tables.cpp
```diff
@@ -8,12 +8,16 @@
...
 #include "calc_tables.hpp"
+#include <algorithm>
+#include <numeric>
 #include <vector>
 
 #include <pbn.hpp>
 #include <solve_board.hpp>
 #include <api/solve_board.hpp>
 #include <solver_if.hpp>
+#include <lookup_tables/lookup_tables.hpp>
```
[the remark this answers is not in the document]
tameware: addressed

library/src/system/parallel_boards.cpp
```diff
@@ -31,23 +31,51 @@ auto resolve_worker_count(
...
   }
 
+  // Map a dispatch slot to the board number to process. With an order, hand out
+  // boards in that sequence (e.g. hardest first); otherwise in index order. The
+  // order is only honored when it is a valid permutation of [0, count); a
+  // malformed order falls back to index order to avoid invalid board indices.
+  const bool use_order =
+    (order != nullptr &&
+     static_cast<int>(order->size()) == count &&
+     is_permutation_of_range(*order, count));
```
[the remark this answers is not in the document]
tameware: Addressed

library/src/system/deal_fanout.cpp
```diff
@@ -0,0 +1,48 @@
...
+    for (int s = 0; s < DDS_SUITS; s++)
+    {
+      c = static_cast<int>(dl.remainCards[h][s] >> 2);
+      fanoutSuit += group_data[c].last_group_ + 1;
+      if (c == 0)
+        numVoids++;
+    }
+    fanoutSuit += numVoids * fanoutSuit;
+    fanout += fanoutSuit;
+  }
```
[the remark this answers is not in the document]
tameware: Addressed. Will look into updating our style guide.

PR #220, opened 2026-06-29
hands: make list1000.txt 1000 unique deals

tameware:
Regenerate list1000.txt from the first 1000 unique deals in list10000.txt (deduplicated by card layout), so the standard 1000-deal benchmark set has no repeats. Preserve the previous duplicate-containing data as list1000_with_dups.txt.


zzcgumn (approved):
Thanks for cleaning this up.

PR #223, opened 2026-07-03
Two percent solution

tameware:
Fixes that give a small performance improvement. Merged from branch opus-two-percent.


tameware:
Replaced by PR https://github.com/dds-bridge/dds/pull/224

PR #229, opened 2026-07-08
Restore scheduler-based calc threading with persistent worker contexts.

tameware:
Honor SetResources thread limits again so batch calc uses strain-aware scheduling and reuses per-slot SolverContexts instead of fresh std::thread workers on every CalcAllTables call.

### DIFF
--- a/library/src/calc_tables.cpp
+++ b/library/src/calc_tables.cpp
@@ -8,6 +8,10 @@
 */
 
 #include "calc_tables.hpp"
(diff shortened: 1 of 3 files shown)


zzcgumn:
Am I right that this is the pull request we will be closing?

tameware:
You are correct. Closing now in favor of PR https://github.com/dds-bridge/dds/pull/216

PR #233, opened 2026-07-14
Ship a JVM-consumable native shared library

### LINKED ISSUE #231: Create a Shared Library Target Suitable for Java Native Interface.

zzcgumn:
Adds build support that produces a single native shared library exporting the stable DDS C ABI, plus a Java Foreign Function & Memory (FFM / Project Panama) binding that calls it — making the solver consumable from the JVM the same way it already is from Python, .NET, and the browser. 

### Layer 1 — the shared library (independently useful; also unblocks .NET/ctypes/JNA)

- Pure-C ABI shim (`library/src/api/dds_c_api.h/.cpp`) wrapping the modern `SolverContext` API with a C-valid, pointer-only, POD-only surface over an opaque `void*` handle — so non-C++ consumers never see `SolverConfig`/`TTKind`/C++ references.
- `//jni:dds_shared cc_shared_library` producing one self-contained `libdds.so` / `libdds.dylib` / `dds.dll` with the whole solver linked in statically.
- Link-time export control: `jni/gen_export_lists.py` generates a Linux version-script and a macOS exported-symbols list from the headers; `//jni/tests:export_set_test` asserts the library exports exactly the 44 public C symbols (39 flat `dll.h` + 5 `dds_c_*`) with zero C++-mangled leakage.

### Layer 2 — JVM consumption

- Adds `rules_java` and a hermetic JDK toolchain (`remotejdk_25`, pinned via `.bazelrc`); JDK ≥ 22 means `java.lang.foreign` is stable, no `--enable-preview`.
- `//jni:dds_ffm` — hand-written `org.dds.ffm` FFM bindings: `struct MemoryLayouts` and `Linker` downcall handles for the shim plus `GetDDSInfo`.
- `//jni:dds_ffm_smoke_test` — loads the built library and solves a known deal through the shim, asserting `cards=1, score[0]=13` (cross-checked against the Python binding).
- `docs/jni_interface.md` — usage guide mirroring the existing per-binding docs.

### Notable design decisions

- No jextract. It ships only as rot-prone early-access binaries (not on BCR, no stable URL/hash), which would make the build non-hermetic. Since it only generates `plain java.lang.foreign`code, the bindings are hand-written instead — the build stays fully hermetic.
- JDK 25, not exactly 22. `rules_java` 9.6.1 registers only `remotejdk_21/_25`; 25 satisfies the ≥22 FFM-stable requirement.
- Handle named `DDS_C_SOLVER_CTX` (not `DDS_SOLVER_CTX`) to avoid a conflicting typedef with `dds_api.hpp`.
- Deferred (non-goals): hand-written JNI convenience API, native-in-JAR/Maven packaging, and switching the whole build to `-fvisibility=hidden` (exports are constrained at link time instead).

## Testing

- `bazel build //jni:dds_shared`
- `bazel test //jni/...         # export_set_test + dds_ffm_smoke_test`
- `bazel test //python/... //web/...   # existing bindings unaffected`
- `Full //... build` and the existing test suites pass. The new tests are scoped off Windows (different export/dumper model and library path).

Closes #231


zzcgumn:
@tzimnoch and @mortensp , I am wondering whether this C API can be used to simplify the .Net integration.

tameware:
I'm not qualified to review this PR. That said, I see Copilot has been busy. If it's happy, I'm happy! Let me know if you'd like me to approve.

zzcgumn:
@tameware , I have to admit that I don't know much about these things either. It seems like Claude Code and Github Copilot are getting somewhere. I would be grateful if you can approve to say that you think this is a good idea, even though you are not qualified to analyse the code.

tameware:
Will do. I’ll also fire off another Copilot review since I still have
quota.

jni/java/org/dds/ffm/Dds.java
```diff
@@ -0,0 +1,298 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
+/**
+ * FFM bindings for the DDS native shared library. Load one instance per library
+ * file with {@link #load(Path)}; it owns an {@link Arena} scoping the native
+ * library and any segments allocated from {@link #arena()}. Not thread-safe:
+ * use one native SolverContext per thread.
+ */
+public final class Dds implements AutoCloseable {
```
tzimnoch: This class should not be final. Let people extend it if they want to.

jni/java/org/dds/ffm/Dds.java
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+
+    /** Destroy a solver context returned by {@link #createSolverContext()}. */
+    public void destroySolverContext(MemorySegment ctx) {
+        invokeVoid(destroyContext, ctx);
+    }
+
+    /** Solve a single board. Returns a RETURN_* status code. */
+    public int solveBoard(MemorySegment ctx, MemorySegment deal, int target,
```
tzimnoch, line 256: The constants defined in dll.h aren't visible here, are they?

Should the RETURN_* error codes be defined as enums on the java side or at least exposed as constants somewhere in the namespace?

jni/java/org/dds/ffm/Dds.java
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    private static String archToken() {
+        String arch = System.getProperty("os.arch", "").toLowerCase(java.util.Locale.ROOT);
+        if (arch.equals("aarch64") || arch.equals("arm64")) {
+            return "aarch64";
+        }
+        if (arch.equals("x86_64") || arch.equals("amd64")) {
+            return "x86_64";
+        }
```
tzimnoch, line 215: For those of us still fond of 8-bit architectures, this wounds my heart a little.

Will this work on a 32-bit JVM or is this 64-bit only? Are you using OpenJDK, Azul, or something else?

It is kind of silly to use a 32-bit JVM on a 64-bit machine, but it does work. Should 64-bit JVM be added to the list of project requirements/prerequisites?

tzimnoch:
> @tzimnoch and @mortensp , I am wondering whether this C API can be used to simplify the .Net integration.

I have no opinion on that.

jni/java/org/dds/ffm/Dds.java
zzcgumn: Initial scope for this work was to make DDS 3.x available to people who use it in Java applications. Is it okay to leave discussions about 32 vs 64 bit JVM to a follow up pull request?

docs/jni_interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+---
+
+## Prerequisites
+
+- **JDK 22 or later.** `java.lang.foreign` is a stable API from JDK 22, so no
+  `--enable-preview` is needed. The Bazel build is hermetic: `rules_java`
+  supplies the JDK (currently pinned to `remotejdk_25` in
+  [`.bazelrc`](../.bazelrc)); you do **not** need a system JDK installed.
+- **Bazel 8.x or later.** (The committed `MODULE.bazel.lock` uses lockFileVersion 28; Bazel 7 users must delete/regenerate the lock file.)
```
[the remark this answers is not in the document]
zzcgumn: Documentation for all targets will be updated in a follow up pull request.

jni/java/org/dds/ffm/Dds.java
tzimnoch: I don't think 32-bit support should be added.

I would call out the 64-bit requirement. This should not work with a JVM in 32-bit mode even if running on a 64-bit machine.

zzcgumn:
I think this is ready to be merged now. @tameware and @tzimnoch, could you please approve if you agree that this pull request can be merged now?

tameware (approved):
Copilot left one more comment. It's pedantic. Fine for me to submit with or without Copilot's suggested change.

PR #234, opened 2026-07-14
Validate card counts in CalcAllTablesN to prevent segfault on malformed deals

danielneill:
## Summary
`calc_all_tables_pbn` / `CalcAllTablesN` can segfault when a batch contains a malformed deal — one whose hands don't each hold 13 cards. The same deal solved alone returns `RETURN_CARD_COUNT` (-14) cleanly.

## Mechanism
A malformed deal passes `convert_from_pbn` and reaches the multi-threaded solver, which reads out of bounds before the per-board card-count check fires. The crash is intermittent — it depends on the number of tables in the batch and the heap layout (a 32-table batch crashes; the same malformed deal in a 2-table batch returns -14).

## Fix
Validate up front in `CalcAllTablesN` that every hand holds exactly 13 cards (`std::popcount` over the four suit bitmasks), returning `RETURN_CARD_COUNT` before any board is dispatched to the solver. This matches the contract the single-board path already enforces.

Scope notes:
- Covers the wrong-count case (the observed crash). Cross-hand duplicate cards (each hand has 13 but a card is repeated) are out of scope.
- The single-board path (`CalcDDtableN`) is unchanged — it already returns -14 without crashing.

## Testing
- New regression test: a malformed deal raises both solo and inside a batch (previously segfaulted the interpreter).
- Verified against a build of this branch: a 32-deal batch that segfaulted (exit 139) now raises -14; valid batches still return tables; all existing `//python:all` tests pass.

### DIFF
--- a/library/src/calc_tables.cpp
+++ b/library/src/calc_tables.cpp
@@ -8,6 +8,7 @@
 */
 
 #include "calc_tables.hpp"
+#include <bit>
 #include <vector>
 
 #include <pbn.hpp>
@@ -253,6 +254,22 @@ int STDCALL CalcAllTablesN(
 
   if (count * dealsp->no_of_tables > MAXNOOFTABLES * DDS_STRAINS)
     return RETURN_TOO_MANY_TABLES;
+
+  // Each hand must hold exactly 13 cards. A malformed deal otherwise reaches
+  // the multi-threaded solver, which can read out of bounds and crash the
+  // process -- intermittently, depending on the number of tables and heap
+  // layout -- before the per-board card-count check fires. Reject up front so
+  // batch calls fail cleanly with RETURN_CARD_COUNT, matching the contract the
+  // single-board solver already enforces.
+  for (int m = 0; m < dealsp->no_of_tables; m++)
+    for (int h = 0; h < DDS_HANDS; h++)
+    {
+      int cardsInHand = 0;
+      for (int s = 0; s < DDS_SUITS; s++)
+        cardsInHand += std::popcount(dealsp->deals[m].cards[h][s]);
+      if (cardsInHand != 13)
+        return RETURN_CARD_COUNT;
+    }
 
   int ind = 0;
   int lastIndex = 0;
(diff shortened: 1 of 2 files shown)


library/src/calc_tables.cpp
```diff
@@ -254,6 +255,22 @@ int STDCALL CalcAllTablesN(
   if (count * dealsp->no_of_tables > MAXNOOFTABLES * DDS_STRAINS)
     return RETURN_TOO_MANY_TABLES;
 
+  // Each hand must hold exactly 13 cards. A malformed deal otherwise reaches
```
zzcgumn, line 258: Are we sure that this is what we want? I think all hands have the same number of cards and at most 13 cards is the correct criteria as partial deals are allowed.

tzimnoch: +1

I also prefer to hide this feature behind a flag. While this isn't enormously expensive, I don't want to take the performance hit to verify something I already guaranteed some other way. This is a speed-critical library for many use cases.

danielneill:
Closing this PR — after a deeper review I can't stand behind its premise.

**@zzcgumn is right about the criterion.** The single-board validator (`board_value_checks`) enforces *equal card counts across the four hands* (and ≤52 total), not 13 per hand — partial deals are legal. My exactly-13 check would have rejected input the library correctly accepts today. That alone would have required rework.

**More importantly, I can no longer reproduce the segfault the PR claims to fix.** I went back with the actual malformed deals from the production data that triggered the original report and tried, with the guard removed:

- C++ ASAN harness (batch sizes 8–40, malformed deal at various positions, ~40 runs)
- `-c opt` Python builds of this branch's parent and of pre-#224 `develop` (~800 calls, including the batch sizes recorded as crashing, sequential multi-batch runs over the real deal corpus, and batches containing multiple malformed deals)
- the exact production wheel binary, on the machine where the crash was observed

Every single call returned a clean `RETURN_CARD_COUNT` (-14) from the existing per-board validation. No crash, no ASAN findings. Tracing the code agrees: `board_range_checks`/`board_value_checks` run before the solver touches anything dangerous, in both the solo and batch paths.

The original SIGSEGV was observed inside a long-running, heavily loaded analysis process solving ~20k deals. Whatever happened there, I can't substantiate that a malformed deal in a batch is the cause — so an upfront guard justified by that claim shouldn't land, and @tzimnoch's reluctance to pay for redundant validation (however small) is fair when the validation is provably redundant in every environment I can test.

If I ever catch the crash again with a traceable stack or ASAN report, I'll come back with evidence and a properly targeted fix. Thanks to all three reviewers for pushing on the right things.

PR #235, opened 2026-07-15
Automatically fill in the fourth hand when three hands have exactly 13 cards each

tameware:
Deployed at https://tameware.com/adam/bridge/dds/dds_mvp.html

### DIFF
--- a/web/dds_mvp.js
+++ b/web/dds_mvp.js
@@ -386,10 +386,6 @@ function inputIsValid(hands) {
     for (const direction of Object.keys(hands)) {
         const hand = hands[direction];
 
-        if (hand.length != 13) {
-            return "Please enter 13 cards per hand.";
-        }
-
         for (const card of hand) {
             const pip = card.substring(1);
 
--- a/web/dds_mvp.js
+++ b/web/dds_mvp.js
@@ -22,6 +22,9 @@
             rotateClockwise
             pageLoad
             sendJSON
+            fourthHandFillState
+            fillFourthHand
+            updateFourthHandFill
             */
 
 // It's also useful to pass the code through
@@ -104,6 +107,8 @@ function fillFormWithTestData(nesw) {
     for (const element of hand_elements()) {
         element.value = holdings.shift();
     }
+
+    updateFourthHandFill();
 }
 
 function fillFormWithGrandSlamTestData() {
@@ -166,6 +171,7 @@ function clearTestData() {
         element.value = "";
     }
 
+    updateFourthHandFill();
     focusNorthSpades();
 }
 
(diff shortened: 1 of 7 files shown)


web/dds_mvp.js
```diff
@@ -187,6 +194,169 @@ function rotateClockwise() {
...
+    const doubleDummyButton = document.getElementById("double-dummy-it");
+
+    if (fillButton) {
+        fillButton.disabled = !fillState.canFill;
+    }
+
+    if (doubleDummyButton) {
+        doubleDummyButton.disabled = !allHandsHaveThirteenCards(hands);
+    }
+}
```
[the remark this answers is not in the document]
tameware: Removed the unreachable validation path. Added a row showing the cards that have and have not been added to the diagram. Added notes for hands that have more than 13 cards.

tameware:
I thought of a better way to do this. Converting to draft…

web/dds_mvp.js
```diff
@@ -187,6 +199,212 @@ function rotateClockwise() {
...
+
+function updateActionButtons() {
+    let hands = collectHands();
+    const fillState = fourthHandFillState(hands);
+    const doubleDummyButton = document.getElementById("double-dummy-it");
+
+    if (fillState.canFill) {
+        applyFourthHandFill(hands, fillState.emptyHand);
+        hands = collectHands();
```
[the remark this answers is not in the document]
tameware: Updated the PR title.

web/dds_mvp.js
```diff
@@ -187,6 +199,212 @@ function rotateClockwise() {
...
+        return;
+    }
+
+    const hands = collectHands();
+
+    if (allHandsHaveThirteenCards(hands)) {
+        event.preventDefault();
+        sendJSON();
+    }
 }
```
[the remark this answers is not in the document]
tameware: Addressed.

web/dds_mvp.js
```diff
@@ -216,10 +487,6 @@ function inputIsValid(hands) {
     for (const direction of Object.keys(hands)) {
         const hand = hands[direction];
 
-        if (hand.length != 13) {
-            return "Please enter 13 cards per hand.";
-        }
-
         for (const card of hand) {
             const pip = card.substring(1);
```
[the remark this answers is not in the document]
tameware: Resolved.

tzimnoch:
Overall this is getting further and further removed from the concept of minimum viable product and seems to be trying to approach a more mature user interface. That said, will comment on implementation.

web/dds_mvp.js
```diff
@@ -47,6 +51,12 @@ const SUIT_SYMBOLS = {
     "D" : "<span style='color: red'>&diams;</span>",
     "C" : "&clubs;"
 };
+const DECK_SUIT_SYMBOLS = {
+    "S" : "&spades;",
+    "H" : "&hearts;",
+    "D" : "&diams;",
+    "C" : "&clubs;"
+};
```
tzimnoch, line 59: Do you really want to sometimes color-code suit symbols and sometimes not?

Eventually a css file should be split out and style='color: red' replaced with something like class='hearts'.

web/dds_mvp.js
tzimnoch, line 44: There's some redundancy here.

Where DIRECTION_LETTERS are used as keys for a map, DIRECTIONS works just as well. Nothing compels the map key to be a single-character string. hands['N'] is not significantly more performant than hands['north'].

It can help other developers not to have to remember if any particular structure is indexed by the letter or by the word by using just one.

web/dds_mvp.js
```diff
@@ -187,6 +211,253 @@ function rotateClockwise() {
...
+
+function fourthHandFillState(hands) {
+    const handCounts = DIRECTION_LETTERS.map((direction) => hands[direction].length);
+    const fullHands = handCounts.filter((count) => count === 13).length;
+    const emptyHands = handCounts.filter((count) => count === 0).length;
+    const partialHands = handCounts.filter((count) => count > 0 && count < 13).length;
+
+    if (fullHands !== 3 || emptyHands !== 1 || partialHands > 0) {
+        return { canFill: false };
+    }
```
tzimnoch, line 298: This feels too constraining. If the user enters one card in west and then populates north, south, and east, you won't autofill the rest of the west hand. But you would if west were left blank.

This could be intentional.

web/dds_mvp.js
```diff
@@ -187,6 +211,253 @@ function rotateClockwise() {
...
+        for (const card of hands[direction]) {
+            const pip = card.substring(1);
+
+            if (!PIPS.includes(pip)) {
+                return { canFill: false };
+            }
+
+            if (usedCards[card]) {
+                return { canFill: false };
+            }
```
tzimnoch, line 317: These are confirming something that should be true by invariance. If there are duplicate or non-bridge cards assigned to a hand then something has really gone wrong in a way that isn't going to be fixed by refusing to deal out the rest of the deck.

web/dds_mvp.js
```diff
@@ -187,6 +211,253 @@ function rotateClockwise() {
...
+    for (const card of cards) {
+        holdings[card.charAt(0)] += card.charAt(1);
+    }
+
+    for (const suit of ["S", "H", "D", "C"]) {
+        holdings[suit] = holdings[suit]
+            .split("")
+            .sort((a, b) => PIPS.indexOf(a) - PIPS.indexOf(b))
+            .join("");
+    }
```
tzimnoch, line 342: Things like this go away (or at least become less awkward) if the hands are represented as arrays of cards rather than a string of pips. This requires multiple changes throughout the implementation.

It's worth thinking through and settling on one representation that meets a broader variety of use cases so that you don't need specialized functions to convert from one representation to another.

tameware:
Agreed, it's becoming more than an MVP. That's in part because I want to use it myself!

Would you suggest moving it to its own repo, @tzimnoch ?

web/dds_mvp.js
```diff
@@ -187,6 +211,253 @@ function rotateClockwise() {
...
+
+function directionName(directionLetter) {
+    return {
+        N: "north",
+        E: "east",
+        S: "south",
+        W: "west"
+    }[directionLetter];
+}
+
```
tzimnoch, line 355: This belongs with the other lookup consts and can be declared as a lambda, but it might be simpler to just use the long-form suit name throughout and not convert between the two.

web/dds_mvp.js
tameware: 1. Thanks for the thorough review!
2. Cursor pushed back, but I convinced it that DECK_SUIT_SYMBOLS is unnecessary.
3. Agreed re CSS. There's a comment in there about custom elements. I'll do something about it in a future PR.

web/dds_mvp.js
```diff
@@ -187,6 +211,253 @@ function rotateClockwise() {
...
         element.value = hands.shift();
     }
+
+    updateActionButtons();
+}
+
+function allDeckCards() {
+    const cards = [];
+
+    for (const suit of ["S", "H", "D", "C"]) {
```
tzimnoch, line 221: ["S", "H", "D", "C"] appears multiple times in source. Similar arrays are declared as constants.

tzimnoch:
I struggle to think of ways to make significant simplifications and clean up, and not for lack of trying the entire exercise myself.

My own project is in React, which allows some modularization of components. This is not an easy exercise.

web/dds_mvp.js
tameware: Done, also for hands and DIR_TO_HAND. This simplifies the code and avoids possibly mistaking _N = NT_ with _N = north_.

tzimnoch:
I shared my private repo. Feel free to have a look around. It's not actually in any cleaner a state, but you might find implementation ideas you're welcome to steal.

If I were less embarrassed by the current quality of it, I would make it public.

The latest build I bothered to push is testable at https://tzimnoch.github.io/cbb.testsite/

web/dds_mvp.js
tameware: Taken care of with prior change.

zzcgumn (approved):
My web knowledge is very limited, I agree that the use case has real value.

web/dds_mvp.js
tameware: Made it a constant and replaced with names rather than letters, per @tzimnoch's previous comment. Also added a "Card" structure to make the code more readable.

web/dds_mvp.js
tameware: We now have a Card structure. This PR is getting big - more cleanup will come in the future. Thanks for the suggestion!

web/dds_mvp.js
tameware: It was intentional, but I prefer your suggestion! Will implement it, then resolve this comment.

tameware: I changed my mind. Suppose all four hands are filled, and the user wants to move two cards. As soon as he removes one it will be replaced. Madness! There are ways around that, but best to leave for a future PR.

web/dds_mvp.js
tameware: Addressed, and addressing located a missing test case, not added.

tameware:
Thanks for sharing your repo, @tzimnoch! I will surely steal some features from in in future PRs.

PR #238, opened 2026-07-17
Agent instructions

tameware:
1. Use snake_case for C and C++ code.
2. Follow TTD.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,6 +1,6 @@
 # Naming
 
-For C and C++ code, use snake_case for functions, methods, variables, and parameters.
+For C and C++ code, use snake_case for functions, methods, variables, and parameters. For C++ types use PascalCase, and for member variables use snake_case with a trailing underscore.
 
 Prefer:
 - `deal_fanout`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,3 +15,17 @@ Avoid:
 Exceptions:
 - Match existing external or legacy APIs (for example, public C API names and types that already use a different style).
 - Do not rename unrelated legacy identifiers in the same change unless the task requires it.
+
+# Test-driven development
+
+For every code change, follow strict TDD:
+
+1. Write or update automated tests that initially fail and describe the desired behavior and edge cases.
+2. Write the minimal production code needed to make the new tests pass. Do not add untested functionality.
+3. After tests are green, refactor tests and implementation for clarity and to remove duplication.
+4. Run the relevant tests after each refactor and keep them green.
+5. Repeat this red-green-refactor cycle in small steps.
+
+Structure tests with Arrange-Act-Assert and use descriptive names that capture intent, including negative and boundary cases.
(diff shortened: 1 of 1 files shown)


zzcgumn:
I have started work to create a set of specifications to help coding agents. I suggest we add a project AGENTS.md file either as part of this upcoming pull request or in a later one. Our `.github/copilot-instructions.md` file does also need to be updated.

tameware:
I forgot that this was still a draft. I've promoted it. I'll merge it now with your approval, @zzcgumn - we can add your stuff to it when you're ready. I needed this because I found that Cursor was ignoring `.github/copilot-instructions.md`.

AGENTS.md
```diff
@@ -0,0 +1,27 @@
+# General
+
+Follow .github/instructions/cpp.instructions.md
+
```
zzcgumn, line 8: I think we should add

```
# Intended audience

These instructions have been created to help Cursor but are expected to be generally helpful.
```

AGENTS.md
```diff
@@ -0,0 +1,27 @@
...
+# Naming
+
+Follow .github/instructions/cpp.instructions.md
+
+Exceptions:
+- Match existing external or legacy APIs (for example, public C API names and types that already use a different style).
+- Do not rename unrelated legacy identifiers in the same change unless the task requires it.
+
+# Test-driven development
+
```
zzcgumn, line 18: I like using test driven development with coding agents but it is not a silver bullet. Maybe add something like this?

```
Apply these instructions when test driven development has been requested. Always ask if it is unclear whether TDD should be used.
```

zzcgumn (changes requested):
Couple of things to discuss about this. I had failed to appreciate that you cannot point Cursor to the `.github` directory for instructions in any other way.

AGENTS.md
tameware: Done.

AGENTS.md
tameware: @zzcgumn > Maybe add something like this?

I've had good results using TDD always. I learned about it in 2000. My code after that was more robust and easier to follow, but only when I remembered to use it! In college in the early '80s we'd been taught that our code should have strong cohesion and weak coupling, but not how to achieve that. It seemed an art. TDD is a tool that almost forces it. 

That said, I did have to add the line that says it should not apply to documentation!

tameware: Kent Beck noted that TDD helps maximize the amount of code not written. This should be useful when we deal with AIs, since they love writing voluminous code.

zzcgumn: I am happy for TDD to be the default mode we ask agents to work with. Beck's point is very interesting as they tend to write lot of code, repeat themselves and write comments that explain what is obvious.

zzcgumn (approved):
We can return to the TDD discussion later. I suppose one advantage is that we try to force all contributors to write well-tested code.

PR #244, opened 2026-07-19
Check createSolverContext for NULL in embedded FFM smoke test.

tameware:
Aligns DdsEmbeddedSmokeTest with DdsSmokeTest so a failed context allocation fails early with a clear assertion instead of a later RETURN_UNKNOWN_FAULT from solveBoard.


zzcgumn (approved):
Good find.

PR #248, opened 2026-07-19
Web

tameware:
Multiple UI improvements.

Deployed to https://tameware.com/adam/bridge/dds/dds_mvp.html


web/tests/dds_mvp_test.mjs
```diff
@@ -255,3 +430,338 @@ test("fillFormWithEveryoneMakes3nTestData populates inputs", () => {
...
+    assert.equal((deckStatus.match(/data-card=/g) ?? []).length, 52);
+    assert.equal((deckStatus.match(/&spades;/g) ?? []).length, 1);
+    assert.equal((deckStatus.match(/&hearts;/g) ?? []).length, 1);
+    assert.equal((deckStatus.match(/&diams;/g) ?? []).length, 1);
+    assert.equal((deckStatus.match(/&clubs;/g) ?? []).length, 1);
+    assert.match(deckStatus, /data-card="SA"/);
+    assert.match(deckStatus, /data-card="C2"/);
+    assert.match(
+        deckStatus,
+        /class="deck-suit-symbol suit-red">&hearts;<\/span>/
```
tzimnoch: Might as well start using a custom tag.

Either of these syntaxes will work.
```
<heart-suit>AK32</heart-suit>
<heart-suit></heart-suit>QT7
```

```
heart-suit {
  color:black;  
}
heart-suit:before {
  content:'\2665';  /* CSS doesn't take html entities */
  color:red;
}
```

web/tests/dds_mvp_test.mjs
```diff
@@ -255,3 +430,338 @@ test("fillFormWithEveryoneMakes3nTestData populates inputs", () => {
...
+    assert.doesNotMatch(
+        deckStatus,
+        /class="deck-card [^"]*red[^"]*" data-card="HA"/
+    );
+});
+
+test("updateActionButtons grays cards entered in any hand", () => {
+    const document = createMockDocument({
+        north_spades: "A",
+        east_hearts: "k",
```
tzimnoch, line 614: If the lower-case k is intentional, it's not captured in the test description. Good that this works.

web/tests/dds_mvp_test.mjs
```diff
@@ -255,3 +430,338 @@ test("fillFormWithEveryoneMakes3nTestData populates inputs", () => {
...
+    assert.match(
+        deckStatus,
+        /class="deck-card" data-card="SK"/
+    );
+});
+
+test("updateActionButtons shows a card-count note for a hand over 13 cards", () => {
+    const document = createMockDocument({
+        north_spades: "AKQJT98765432A",
+    });
```
tzimnoch, line 639: This test case should fail for a different reason. When this document is loaded the UI should refuse to accept the north hand.

I'd spread the cards out so that this test remains valid if/when there's a change in behavior to refuse to accept invalid hands.

web/tests/dds_mvp_test.mjs
```diff
@@ -255,3 +430,338 @@ test("fillFormWithEveryoneMakes3nTestData populates inputs", () => {
...
+test("updateActionButtons hides the card-count note at the 13-card boundary", () => {
+    const document = createMockDocument({
+        north_spades: "AKQJT98765432A",
+    });
+    const ctx = loadDdsMvp(document);
+    ctx.updateActionButtons();
+    document.setValue("north_spades", "AKQJT98765432");
+
+    ctx.updateActionButtons();
+
```
tzimnoch, line 660: Why is the test set up this way? Shouldn't the mock document have the 13-card hand and lines 655-657 be removed?

What do these additional lines test?

web/tests/dds_mvp_test.mjs
tameware: Done.

web/tests/dds_mvp_test.mjs
tameware: I've added a comment.

Per Cursor:

Those lines test the transition back from overflow to a legal hand, not a fresh 13-card load.

Start with 14 cards → first updateActionButtons() shows "14 cards" and unhides the note (same path as the test above).
Edit down to exactly 13 → second call must hide the note and clear its text.
That covers a real sticky-UI bug: show on overflow, then forget to clear/hidden = true when the count drops to ≤13.

If the mock started at 13 and you dropped 655–657, you’d only check the ≤13 branch on a still-empty note. You would not prove that a previously visible "14 cards" note gets torn down.

So the setup is intentional; the name (“…at the 13-card boundary”) is about crossing that boundary, not starting on it.

web/tests/dds_mvp_test.mjs
tameware: Updated the name.

web/tests/dds_mvp_test.mjs
tameware: I'm not sure what you mean. The UI accepts hands with 14+ cards - perhaps it should not. Or would you like to guarantee that no hand contains more than 13 cards before it's sent to dds?

tameware:
Thanks, @tzimnoch, for another thorough review!

web/tests/dds_mvp_test.mjs
tzimnoch: The UI accepts 14 card hands, but it might not always accept duplicate cards. You can enforce that various ways. What I did if the user tried to add a card that already existed in another player's hand was to move the card from where it was to where the user wants to put it. If the user tried to add a duplicate card, then it would just do nothing.

You're not doing anything like this yet, but you very well could.

So I'd have the test give a 14-card hand spread over multiple suits such that the 14-card hand at least came from a valid deck.

web/tests/test_dds_mvp_css.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class DdsMvpCssTest(unittest.TestCase):
+    def test_deck_status_displays_all_cards_in_one_row(self) -> None:
+        css = (WEB_ROOT / "dds_mvp.css").read_text(encoding="utf-8")
+        body = _rule_body_for_selector(css, "#deck-status")
+        self.assertRegex(body, r"display\s*:\s*flex")
+        self.assertRegex(body, r"flex-wrap\s*:\s*nowrap")
+        self.assertRegex(body, r"font-size\s*:\s*calc\(1em\s*\+\s*1pt\)")
+        self.assertNotRegex(body, r"overflow(?:-x)?\s*:\s*(?:auto|scroll)")
```
tzimnoch: Comment isn't specific to these lines. Be aware there's a limitation to this approach.

This only verifies that the rules match between the css file and the test file, i.e., that you've defined the rule twice and they match. It doesn't test that the rules are applied the way you wish. Rules with higher specificity override rules with lower specificity.

To truly test css you need to render the page and inspect the computed styles on the elements.

web/tests/test_mvp_e2e.py
```diff
@@ -171,6 +171,40 @@ def test_page_load_shows_valid_pips(self) -> None:
...
+            self.assertEqual(
+                page.locator('#deck-status [data-card="SA"]').get_attribute("class"),
+                "deck-card",
+            )
+
+            page.locator("#north_spades").fill("A")
+
+            entered_card = page.locator('#deck-status [data-card="SA"]')
+            self.assertIn("deck-card-entered", entered_card.get_attribute("class"))
+            self.assertLess(float(entered_card.evaluate("el => getComputedStyle(el).opacity")), 1.0)
```
tzimnoch: ok, so you do have this form of css test.

It's really the only test that's valuable. The other css tests should be replaced with this pattern. It's also a lot easier to read than various regex expressions.

web/dds_mvp.css
```diff
@@ -69,6 +69,45 @@
...
+
+/* Bold outline only when Enter would activate Double-dummy it! */
+#double-dummy-it.default-action {
+  outline: 3px solid #222;
+  outline-offset: 2px;
+}
+
+#double-dummy-it:disabled {
+  opacity: 0.45;
+  color: #777;
```
tzimnoch, line 118: AI code bots made the same mistake in my code base.

Opacity is the right style choice for things that should be see through, like a card you're dragging across the UI. You want to see what's under the card.

For disabled or grayed out text, only the color should be changed. If you decide to go with a green felt background instead of white, you don't want this component to become partially green when disabled.

web/tests/test_mvp_e2e.py
tameware: Agreed! Done. The disadvantage of this approach is that the e2e tests are slow enough that we don't include them in test /..., so anyone working on the front end needs to remember to run them manually. I'll look into adding them to the CI, hopefully in a way that won't slow it down substantially.

tameware: Added Issue https://github.com/dds-bridge/dds/issues/256

web/tests/test_dds_mvp_css.py
tameware: Addressed.

web/dds_mvp.css
tameware: Nice catch. Done.

web/tests/dds_mvp_test.mjs
tameware: I see. So the test doesn't fail for a reason other that what it's attempting to test.

Done, and similarly done for `updateActionButtons hides the card-count note at the 13-card boundary()`.

web/dds_mvp.css
```diff
@@ -69,6 +69,65 @@
...
+spade-suit,
+heart-suit,
+diamond-suit,
+club-suit {
+  color: #000;
+}
+
+spade-suit:before {
+  content: '\2660';
+}
```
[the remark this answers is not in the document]
tameware: Done.

PR #250, opened 2026-07-19
Add C++20 likely/unlikely hints on AB and quick-tricks hot paths.

tameware:
Guide the compiler toward fall-through search and common QT suit cases so rare void/cutoff branches mispredict less often.


tameware:
Copilot opened its own PR to fix this one.

PR #257, opened 2026-07-22
Ignore bazel shutdown failures in Windows CI retries.

tameware:
Match the WASM workflow so a missing server cannot abort the PowerShell retry loop.


zzcgumn (approved):
Good find.

PR #268, opened 2026-07-24
Refactor

tameware:
Factor out duplication. Each change is a separate commit.

Reviewers, let me know if you'd prefer to see this split into four PRs.


zzcgumn (approved):
This makes sense to me. I like that this pull request improves the unit test coverage.

PR #271, opened 2026-07-24
Declare UTF-8 in the MVP HTML so suit glyphs render correctly.

tameware:
Servers that omit charset from Content-Type were decoding ♠♥♦♣ as Latin-1 mojibake; pin encoding in <head> and cover it with tests.

### DIFF
--- a/web/BUILD.bazel
+++ b/web/BUILD.bazel
@@ -143,6 +143,7 @@ py_test(
 test_suite(
     name = "web_tests",
     tests = [
+        ":dds_mvp_html_test",
         ":dds_mvp_js_test",
         ":dds_mvp_wasm_test",
         ":wasm_scripts_test",
--- a/web/BUILD.bazel
+++ b/web/BUILD.bazel
@@ -75,6 +75,14 @@ py_test(
 )
 
 py_test(
+    name = "dds_mvp_html_test",
+    size = "small",
+    main = "tests/test_mvp_html.py",
+    srcs = ["tests/test_mvp_html.py"],
+    data = ["dds_mvp.html"],
+)
+
+py_test(
     name = "dds_mvp_js_test",
     size = "small",
     timeout = "short",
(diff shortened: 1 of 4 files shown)


web/BUILD.bazel
```diff
@@ -74,6 +74,14 @@ py_test(
     ],
 )
 
+py_test(
+    name = "dds_mvp_html_test",
+    size = "small",
+    main = "tests/test_mvp_html.py",
+    srcs = ["tests/test_mvp_html.py"],
```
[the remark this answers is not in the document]
tameware: Fixed.

tameware:
I think I assigned this to @tzimnoch instead of requesting a review. I'm on my phone now and requested a review but can't figure out how to reassign to me. Likely not important!

web/tests/test_mvp_html.py
```diff
@@ -0,0 +1,43 @@
...
+import unittest
+from pathlib import Path
+
+WEB_ROOT = Path(__file__).resolve().parents[1]
+HTML_PATH = WEB_ROOT / "dds_mvp.html"
+
+
+class DdsMvpHtmlCharsetTest(unittest.TestCase):
+    def test_declares_utf8_charset_within_first_1024_bytes(self) -> None:
+        # Browsers only honor <meta charset> in the first 1024 bytes. Without it,
```
tzimnoch, line 14: I also prefer this to using the UTF-8 BOM -- easier to see as plaintext, less likely to be deleted by a future developer who doesn't recognize what it is.

Just want to make sure you're aware of the other option. The main advantage of the BOM is that nothing comes before it, so you don't need to worry about this limitation.

Overall I agree this is the better solve.

tzimnoch (approved):

web/tests/test_mvp_html.py
tameware: I had to look up UTF-8 BOM. Now that I know what it is, I certainly agree! Thanks for the review.

PR #277, opened 2026-07-29
docs: recommend bazelisk for Python tests instead of raw pytest

wopdevries:
The Running Unit Tests section recommended raw `pytest` with manual 
`PYTHONPATH` setup, which fails on Python 3.12 due to missing `iniconfig` 
and `PyThreadState_GetUnchecked` symbol errors when loading `_dds3.so`.

Replace with `bazelisk test //python/...` which handles the Python 
extension path automatically and works out of the box.

Tested on Linux (Python 3.12, Ubuntu 24.04).


zzcgumn (approved):
Good spot.

PR #278, opened 2026-07-29
Cap WASM parallel workers under a heap budget, Emscripten-only.

tameware:
Do not request more than 2GB RAM from WASM, its limit. The messages generated upon exceeding it are apparently harmless, but best to avoid them. 

```
$ bazel run //wasm:run_dtest_wasm -- -f hands/list1000.txt -s calc
```

…
```
Cannot enlarge memory, requested 2150969344 bytes, but the limit is 2147483648 bytes!
```
…

Keep the clamp behind __EMSCRIPTEN__ so native resolve_worker_count stays uncapped.
```

### DIFF
--- a/library/src/system/parallel_boards.cpp
+++ b/library/src/system/parallel_boards.cpp
@@ -18,6 +18,9 @@
 #include <thread>
 #include <vector>
 
+#if defined(__EMSCRIPTEN__)
+#include <api/dds.h>
+#endif
 #include <api/dll.h>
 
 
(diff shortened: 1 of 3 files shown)


zzcgumn (approved):
Maybe we will get a request to do something similar on a raspberry pi.

tameware:
> Maybe we will get a request to do something similar on a raspberry pi.

We might! I added issue https://github.com/dds-bridge/dds/issues/279 to generalize this fix.

PR #283, opened 2026-08-03
Web enhancements

tameware:
On the way to allowing card-by-card play. Stopping for a review before the PR gets even larger. This change is intended to be merged as is, since it does enhance functionality.

Deployed at https://tameware.com/adam/bridge/dds/dds_mvp.html

### DIFF
--- a/specs/web-mvp.md
+++ b/specs/web-mvp.md
@@ -46,7 +46,7 @@ a full application.
   `python3 web/serve_mvp.py` (not plain `http.server`). `file://` remains useful
   for UI-only checks; instantiating the solver module for a solve needs HTTP +
   those headers on any host. Guarded by `dds_mvp_e2e_test` (`test_http_is_cross_origin_isolated`,
-  HTTP part-score / Enter-to-solve).
+  HTTP part-score solve / auto-filled DD table).
 - **Three test tiers, with suite membership as wired in BUILD:**
   - **Unit / JS** (`web_tests`): `dds_mvp_wasm_test` (native `cc_test` over the
     solve logic with `DDS_MVP_WASM_NO_MAIN`), `dds_mvp_js_test` (Node runs
--- a/specs/web-mvp.md
+++ b/specs/web-mvp.md
@@ -45,7 +45,7 @@ a full application.
   `Cross-Origin-Embedder-Policy: require-corp` are set. Serve with
   `python3 web/serve_mvp.py` (not plain `http.server`). `file://` remains useful
   for UI-only checks; instantiating the solver module for a solve needs HTTP +
-  those headers. Guarded by `dds_mvp_e2e_test` (`test_http_is_cross_origin_isolated`,
+  those headers on any host. Guarded by `dds_mvp_e2e_test` (`test_http_is_cross_origin_isolated`,
   HTTP part-score / Enter-to-solve).
 - **Three test tiers, with suite membership as wired in BUILD:**
   - **Unit / JS** (`web_tests`): `dds_mvp_wasm_test` (native `cc_test` over the
--- a/web/BUILD.bazel
+++ b/web/BUILD.bazel
@@ -9,7 +9,7 @@ load("//:wasm_compat.bzl", "WASM_LINKOPTS")
 WASM_MVP_LINKOPTS = WASM_LINKOPTS + [
     "-sMODULARIZE=1",
     "-sEXPORT_NAME=createDdsModule",
-    "-sEXPORTED_FUNCTIONS=['_dds_mvp_calc_table','_malloc','_free']",
+    "-sEXPORTED_FUNCTIONS=['_dds_mvp_calc_table','_dds_mvp_solve_leads','_malloc','_free']",
     "-sEXPORTED_RUNTIME_METHODS=['ccall','getValue']",
     "-sENVIRONMENT=web,worker,node",
 ]
--- a/web/BUILD.bazel
+++ b/web/BUILD.bazel
@@ -117,11 +117,12 @@ py_test(
     deps = [":mvp_site"],
 )
 
-# Playwright usually finishes in ~30s; first run may download Chromium (requires-network).
+# Playwright usually finishes in ~30–45s; first run may download Chromium
+# (requires-network). Full suite can approach 60s under load, so allow moderate.
 py_test(
     name = "dds_mvp_e2e_test",
-    size = "small",
-    timeout = "short",
+    size = "medium",
+    timeout = "moderate",
     main = "tests/test_mvp_e2e.py",
     srcs = ["tests/test_mvp_e2e.py"],
     data = [
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -17,7 +17,7 @@
     "nw north ne"
     "west center east"
     "sw south se";
-  grid-template-columns: 1.5fr 1.5fr 1.6fr;
+  grid-template-columns: 1.5fr 1.5fr 1.5fr;
   background-color: #2196F3;
 }
 
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -65,6 +65,13 @@
   text-align: center;
 }
 
+.grid-item.hand-north,
+.grid-item.hand-east,
+.grid-item.hand-south,
+.grid-item.hand-west {
+  text-align: left;
+}
+
 .grid-filler {
   padding: 20px;
 }
@@ -72,6 +79,50 @@
 .hand-card-count {
   font-size: 1rem;
   margin-top: 0.25rem;
+}
+
+.hand-holding {
+  display: inline;
+}
+
+.hand-suit {
+  cursor: text;
+  display: inline-block;
+  text-align: left;
+  white-space: nowrap;
+}
+
+.hand-suit-input {
+  background: transparent;
+  border: none;
+  caret-color: #111;
+  color: transparent;
+  font: inherit;
+  margin: 0;
+  min-width: 0.65em;
+  outline: none;
+  padding: 0;
+  width: 0.65em;
+}
+
+.hand-suit-input:focus {
+  caret-color: #111;
+}
+
+.hand-card {
+  background: transparent;
+  border: none;
+  color: inherit;
+  cursor: pointer;
+  font: inherit;
+  margin: 0;
+  padding: 0 0.05em;
+}
+
+.hand-card:hover,
+.hand-card:focus-visible {
+  text-decoration: underline;
+  outline: none;
 }
 
 td {
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -236,13 +236,3 @@ diamond-suit {
 .deck-card-entered {
   color: #888;
 }
-
-/* Bold outline only when Enter would activate Double-dummy it! */
-#double-dummy-it.default-action {
-  outline: 3px solid #222;
-  outline-offset: 2px;
-}
-
-#double-dummy-it:disabled {
-  color: #777;
-}
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -143,6 +143,16 @@
   font: inherit;
   margin: 0 0.04em;
   padding: 0 0.12em;
+  position: relative;
+}
+
+.hand-card-tricks {
+  bottom: 0;
+  font-size: 0.38em;
+  font-weight: 700;
+  line-height: 1;
+  position: absolute;
+  right: 0.05em;
 }
 
 .hand-card:hover,
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -78,6 +78,19 @@
   padding: 20px;
 }
 
+/* After .grid-item so font-size wins over the hand diagram's 30px. */
+.grid-item.grid-filler-se {
+  align-items: center;
+  display: flex;
+  font-size: 1rem;
+  justify-content: center;
+}
+
+#result-table {
+  border-collapse: collapse;
+  width: 100%;
+}
+
 .hand-card-count {
   font-size: 1rem;
   margin-top: 0.25rem;
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -89,8 +89,16 @@
 .grid-item.grid-filler-se {
   align-items: center;
   display: flex;
+  flex-direction: column;
   font-size: 1rem;
   justify-content: center;
+}
+
+.result-table-hint {
+  font-size: 0.85rem;
+  line-height: 1.2;
+  margin-bottom: 0.35em;
+  text-align: center;
 }
 
 .contract-status-panel {
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -105,7 +105,12 @@
   font-size: 1.6em;
 }
 
+.contract-status-by {
+  font-size: 1em;
+}
+
 .contract-status-declarer {
+  font-size: 1.6em;
   font-weight: 700;
 }
 
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -8,7 +8,7 @@
 .grid-outer {
   background-color: #2196F3;
   padding: 10px;
-  max-width: 800px;
+  max-width: 1100px;
 }
 
 .grid-container {
@@ -17,7 +17,7 @@
     "nw north ne"
     "west center east"
     "sw south se";
-  grid-template-columns: 1fr 1fr 1fr;
+  grid-template-columns: 1.5fr 1.5fr 1.6fr;
   background-color: #2196F3;
 }
 
@@ -60,6 +60,8 @@
 .grid-item {
   background-color: rgba(255, 255, 255, 0.8);
   border: 0px solid rgba(0, 0, 0, 0.8);
+  /* Allow long holdings to overflow the cell instead of widening the column. */
+  min-width: 0;
   padding: 20px;
   font-size: 30px;
   text-align: center;
@@ -95,7 +97,7 @@
 .hand-suit-input {
   background: transparent;
   border: none;
-  caret-color: #111;
+  caret-color: transparent;
   color: transparent;
   font: inherit;
   margin: 0;
@@ -106,23 +108,41 @@
 }
 
 .hand-suit-input:focus {
-  caret-color: #111;
+  caret-color: transparent;
 }
 
 .hand-card {
   background: transparent;
-  border: none;
+  border: 1px solid #bbb;
+  border-radius: 0.15em;
   color: inherit;
-  cursor: pointer;
+  cursor: text;
   font: inherit;
-  margin: 0;
-  padding: 0 0.05em;
+  margin: 0 0.04em;
+  padding: 0 0.12em;
 }
 
 .hand-card:hover,
 .hand-card:focus-visible {
-  text-decoration: underline;
+  border-color: #555;
   outline: none;
+  text-decoration: none;
+}
+
+.hand-caret {
+  animation: hand-caret-blink 1s step-end infinite;
+  background: #111;
+  display: inline-block;
+  height: 0.85em;
+  margin: 0 0.02em;
+  vertical-align: middle;
+  width: 1px;
+}
+
+@keyframes hand-caret-blink {
+  50% {
+    opacity: 0;
+  }
 }
 
 td {
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -91,6 +91,16 @@
   width: 100%;
 }
 
+#result-table td {
+  cursor: pointer;
+}
+
+#result-table td.result-cell-selected {
+  background-color: #bbdefb;
+  outline: 2px solid #1565c0;
+  outline-offset: -2px;
+}
+
 .hand-card-count {
   font-size: 1rem;
   margin-top: 0.25rem;
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -79,11 +79,34 @@
 }
 
 /* After .grid-item so font-size wins over the hand diagram's 30px. */
+.grid-item.grid-filler-ne {
+  align-items: center;
+  display: flex;
+  font-size: 1.25rem;
+  justify-content: center;
+}
+
 .grid-item.grid-filler-se {
   align-items: center;
   display: flex;
   font-size: 1rem;
   justify-content: center;
+}
+
+.contract-status-panel {
+  align-items: baseline;
+  display: flex;
+  gap: 0.2em;
+  line-height: 1.2;
+  text-align: center;
+}
+
+.contract-status-denom {
+  font-size: 1.6em;
+}
+
+.contract-status-declarer {
+  font-weight: 700;
 }
 
 #result-table {
--- a/web/dds_mvp.css
+++ b/web/dds_mvp.css
@@ -131,6 +131,11 @@
   cursor: pointer;
 }
 
+#result-table td:focus-visible {
+  outline: 2px solid #1565c0;
+  outline-offset: -2px;
+}
+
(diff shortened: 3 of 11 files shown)


web/dds_mvp.js
```diff
@@ -303,6 +395,593 @@ function updateDeckStatus(hands) {
...
+
+    return "<button type=\"button\" class=\"" + classes + "\"" +
+        " data-direction=\"" + direction + "\"" +
+        " data-card=\"" + card.key() + "\"" +
+        " data-index=\"" + index + "\"" +
+        " aria-label=\"" + handCardAriaLabel(direction, card) + "\">" +
+        card.pip +
+        badge +
+        "</button>";
+}
```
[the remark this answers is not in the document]
tameware: Addressed.

web/dds_mvp_wasm.cpp
```diff
@@ -16,11 +18,40 @@
...
+{
+  int n = 0;
+  for (int i = 0; i < fut.cards; ++i) {
+    append_lead_card(out_leads, n, fut.suit[i], fut.rank[i], fut.score[i]);
+    // equals uses Holding encoding: bit r set means rank r is equivalent.
+    for (int rank = 2; rank <= 14; ++rank) {
+      if ((fut.equals[i] & (1 << rank)) != 0) {
+        append_lead_card(out_leads, n, fut.suit[i], rank, fut.score[i]);
+      }
+    }
```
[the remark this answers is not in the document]
tameware: Addressed.

specs/web-mvp.md
```diff
@@ -45,7 +45,7 @@ a full application.
   `Cross-Origin-Embedder-Policy: require-corp` are set. Serve with
   `python3 web/serve_mvp.py` (not plain `http.server`). `file://` remains useful
   for UI-only checks; instantiating the solver module for a solve needs HTTP +
-  those headers. Guarded by `dds_mvp_e2e_test` (`test_http_is_cross_origin_isolated`,
+  those headers on any host. Guarded by `dds_mvp_e2e_test` (`test_http_is_cross_origin_isolated`,
   HTTP part-score / Enter-to-solve).
```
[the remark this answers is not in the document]
tameware: Addressed.

web/dds_mvp.js
```diff
@@ -303,6 +395,589 @@ function updateDeckStatus(hands) {
...
+    onContractSelect(direction, denomination);
+    void scheduleDealSolve();
+}
+
+function clearResultCellSelection() {
+    clearResultCellSelectionHighlight();
+    selectedContractState = null;
+    updateContractStatus();
+    onContractSelect(null, null);
+    void scheduleDealSolve();
```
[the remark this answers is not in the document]
tameware: Addressed.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +18,56 @@
...
+  return ctx;
+}
+
+void append_lead_card(int* out_leads, int& n, int suit, int rank, int score)
+{
+  out_leads[1 + 3 * n] = suit;
+  out_leads[1 + 3 * n + 1] = rank;
+  out_leads[1 + 3 * n + 2] = score;
+  ++n;
+}
```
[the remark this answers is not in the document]
tameware: Addressed in 48f3a0d: expanded leads are hard-capped at 13 with a unit test for the overflow case.

tameware:
Addressed the latest Copilot notes in 8b6822e:

- Clear cached lead-trick badges immediately on hand edit (and invalidate in-flight lead solves).
- Add `aria-label` on each concealed suit input (e.g. "North spades").

tameware:
Addressed the contract-switch lead-tricks race in $(git rev-parse --short HEAD):

Selecting a new DD-table cell now invalidates in-flight lead solves and clears `leadTricksByCardKey` immediately (same pattern as clearing the selection), so numerals from the previous contract cannot linger or repaint from a queued solve.

tameware:
Re the suggestion to change \`td().nth(4)\` to \`nth(5)\` for South/NT: that mixes up two indexing schemes.

Playwright’s \`locator("td").nth(...)\` counts only \`<td>\` cells (C=0 … S=3, **NT=4**). DOM \`row.cells[5]\` is the same cell because \`cells\` also includes the direction \`<th>\` at index 0. \`nth(5)\` would miss the cell.

Left the locators as \`nth(4)\` and asserted \`selectedContract()\` is South/NT after the click in $(git rev-parse --short HEAD).

tameware:
Addressed the \`solveOpeningLeadTricks\` buffer-walk concern in $(git rev-parse --short HEAD):

After reading \`n\` from WASM, reject counts outside \`0..13\` before iterating so a corrupted length cannot read past the fixed 13-lead allocation.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +18,71 @@
...
 namespace {
-auto mvp_context() -> SolverContext& {
-  static SolverContext ctx;   // lazily constructed on first call, lives for
-                               // the module instance's lifetime (the session)
+// Separate contexts so CalcDDtable worker pools cannot disturb SolveBoard
+// (and vice versa) when the UI auto-fills the table then analyzes leads.
+auto mvp_table_context() -> SolverContext& {
+  static SolverContext ctx;
   return ctx;
 }
```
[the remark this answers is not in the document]
tameware: Done.

tameware:
Addressed the result-table keyboard/AT gap in $(git rev-parse --short HEAD):

Data cells now get \`tabIndex=0\`, \`role="button"\`, and contract \`aria-label\`s on load. Enter/Space activate the same selection path as click (Space prevents scroll), with a \`:focus-visible\` outline for keyboard focus.

web/dds_mvp.css
```diff
@@ -17,7 +17,7 @@
     "nw north ne"
     "west center east"
     "sw south se";
-  grid-template-columns: 1fr 1fr 1fr;
+  grid-template-columns: 1.5fr 1.5fr 1.6fr;
```
tzimnoch: Is the additional width on the last column intentional? This is a highly unusual and specific ratio.

web/dds_mvp.html
```diff
@@ -16,7 +16,7 @@
 <html lang="en">
 <head>
     <meta charset="utf-8">
-    <link rel="stylesheet" type="text/css" href="dds_mvp.css">
+    <link rel="stylesheet" type="text/css" href="dds_mvp.css?v=lead-tricks-2">
```
tzimnoch: The param argument looks extraneous.

web/dds_mvp.html
```diff
@@ -78,61 +127,10 @@ <h1>
...
-            <td valign="top"><br></td>
-            <td valign="top"><br></td>
-            <td valign="top"><br></td>
-        </tr>
-    </table>
-
     <script src="dds_mvp_wasm_bin.js"></script>
     <script src="dds_mvp_wasm.js"></script>
-    <script src="dds_mvp.js"></script>
+    <script src="dds_mvp.js?v=lead-tricks-2"></script>
```
tzimnoch: The param argument looks extraneous.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +19,99 @@
...
+void append_lead_card(
+    int* out_leads, int& n, int capacity, int suit, int rank, int score)
+{
+  if (n >= capacity) {
+    return;
+  }
+  out_leads[1 + 3 * n] = suit;
+  out_leads[1 + 3 * n + 1] = rank;
+  out_leads[1 + 3 * n + 2] = score;
+  ++n;
```
tzimnoch: This comes across as a puzzle to solve.

The test on lines 50-52 is already done on lines 63-65 below. You don't need to pass the capacity argument or perform the check here.

++n at the end is very unkind to future maintainers of the code. They expect that you're writing to out_leads. How and why n is getting incremented shouldn't be hidden here. There are ways to get around needing this at all.

tameware:
Thanks for the speedy review, @tzimnoch! I'll get right on these comments.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +19,99 @@
...
+}
+
+void write_expanded_leads(const FutureTricks& fut, int* out_leads)
+{
+  int n = 0;
+  for (int i = 0; i < fut.cards; ++i) {
+    if (n >= kMaxExpandedLeads) {
+      break;
+    }
```
tzimnoch: Doesn't FutureTricks contain up-to 13 entries, one for each possible card the player could lead? Why is the n variable needed at all? Isn't fut.field[i] being written to out_leads[i*3 + 1 + <field_offset>]?

Have I misunderstood what's going on?

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +19,99 @@
...
+      if (n >= kMaxExpandedLeads) {
+        break;
+      }
+      if ((equals_ranks & (1 << (rank - 2))) != 0) {
+        append_lead_card(
+            out_leads, n, kMaxExpandedLeads, fut.suit[i], rank, fut.score[i]);
+      }
+    }
+  }
+  out_leads[0] = n;
```
tzimnoch, line 92: Shouldn't n = fut.cards(.length) at this point?

web/dds_mvp.css
tameware: Not intentional — equalized to `1.5fr 1.5fr 1.5fr`. The slightly wider third column was accidental; we still keep the wider-than-`1fr` columns (plus `min-width: 0` / larger `max-width`) so typical 8-card suits fit without jumping the layout.

web/dds_mvp.html
tameware: Agreed — removed the `?v=` cache-buster. It was only there for local hard-refresh during iteration.

web/dds_mvp.html
tameware: Same as above — dropped the `?v=` query param on the script tag.

web/dds_mvp_wasm.cpp
tameware: Agreed. Replaced `append_lead_card` with a plain `write_lead_triple(out, slot, …)` that only writes; the caller owns the capacity check and does `++n` next to each write.

web/dds_mvp_wasm.cpp
tameware: You understood the packed shape correctly as the common case, but `FutureTricks` is not one entry per leadable card: each of `fut.cards` is a representative rank plus an `equals` bitmask of other ranks that share the same score. We expand those into one `(suit,rank,score)` triple per physical card so the UI can badge every pip — hence a separate `n` that can exceed `fut.cards` (still capped at 13). Clarified that in the comments around `write_expanded_leads`.

web/dds_mvp_wasm.cpp
tameware: Only if there were no equals to expand. After expansion `n` is the number of physical lead cards written (representative + equals ranks), which can be greater than `fut.cards` — so we keep counting writes rather than assigning `fut.cards`.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +19,99 @@
...
+    // rank r is present when bit (r - 2) is set in (equals >> 2).
+    const int equals_ranks = fut.equals[i] >> 2;
+    for (int rank = 2; rank <= 14; ++rank) {
+      if (n >= kMaxExpandedLeads) {
+        break;
+      }
+      if ((equals_ranks & (1 << (rank - 2))) != 0) {
+        append_lead_card(
+            out_leads, n, kMaxExpandedLeads, fut.suit[i], rank, fut.score[i]);
+      }
```
tzimnoch: It is unclear what equals_rank is trying to do, but if the condition on line 75 is ever true then you're going to write one FutureTricks entry to out_leads twice and eventually break out of this function before visiting every FutureTricks result.

Again, I might have misunderstood what's going on. The code from 45-82 largely has me stumped.

tameware: Fair — that block was too dense. Clarified the `equals` contract and tightened the loop:

- Per dll-description, each `fut.cards` entry is the **highest** card of an equivalence group; `fut.equals[i]` lists only **lower** equivalent ranks (Holding encoding: `sequence << 2`), not a second copy of `fut.rank[i]`.
- Expansion now walks `rank` in `[2, representative)` only, so the capacity check cannot “rewrite” the representative, and a stray self/higher bit in `equals` is ignored.
- Hitting the 13-card cap stops further expansion by design (one full hand on lead); remaining `FutureTricks` groups are intentionally skipped once the buffer is full.

Also added a unit test that equals bits at/above the representative do not duplicate it.

web/BUILD.bazel
```diff
@@ -9,7 +9,7 @@ load("//:wasm_compat.bzl", "WASM_LINKOPTS")
 WASM_MVP_LINKOPTS = WASM_LINKOPTS + [
     "-sMODULARIZE=1",
     "-sEXPORT_NAME=createDdsModule",
-    "-sEXPORTED_FUNCTIONS=['_dds_mvp_calc_table','_malloc','_free']",
+    "-sEXPORTED_FUNCTIONS=['_dds_mvp_calc_table','_dds_mvp_solve_leads','_malloc','_free']",
```
tzimnoch, line 12: DDS's API should be exported directly for clients to call with their own wrappers or no wrapper at all if they choose.

"_SetMaxThreads","_SolveBoard","_SolveBoardPBN","_CalcDDtable","_CalcDDtablePBN"

web/BUILD.bazel
```diff
@@ -9,7 +9,7 @@ load("//:wasm_compat.bzl", "WASM_LINKOPTS")
...
     "-sEXPORTED_RUNTIME_METHODS=['ccall','getValue']",
```
tzimnoch, line 13: In my own project I also export "cwrap","setValue", though I'm not sure if that's truly necessary. I would need to revisit this.

tameware: Agreed it isn’t necessary here. This MVP only uses `ccall` (string PBN in, status out) and `getValue` (read the flat `int` buffers). We never write into WASM memory from JS, so `setValue` isn’t needed, and `cwrap` would just wrap the same `ccall` sites. Leaving `EXPORTED_RUNTIME_METHODS` at `ccall`/`getValue`.

web/BUILD.bazel
tameware: Leaving the MVP surface as the thin bridge only (`_dds_mvp_calc_table` / `_dds_mvp_solve_leads`). Those package the PBN → flat-buffer shape the page needs (including equals expansion for lead badges) and own the session `SolverContext`s.

Exporting the full legacy C entry points (`SolveBoard*`, `CalcDDtable*`, `SetMaxThreads`, …) would be a separate WASM binding product — out of scope for this MVP page, and a different ABI than the context-based C++ API the library is moving toward. Happy to discuss a dedicated WASM API target later if you want that shape for your project.

tameware:
Addressed the \`scheduleDealSolve\` queue buildup in $(git rev-parse --short HEAD):

Rapid schedules now only bump \`dealSolveEpoch\`. A single in-flight queue job loops until the epoch stabilizes, so intermediate typing/contract clicks no longer each append a stale promise-chain callback.

web/dds_mvp_wasm.cpp
tzimnoch: Ok, I forgot about this case. If your hand has adjacent cards, like the 4 and 5 of clubs, it doesn't return separate results for each card.

tameware:
I can't find this comment in context, but it sounds as though you're happy
now. Yes, calling only once for equivalent cards (here, that means
adjacent) is an important optimization. DDS does it internally during a
solve, but we're requesting a separate solve for each possible lead.

I've pushed the latest build to tameware.com , and it looks good to me. I
want to increase the size of the pips, but that can come in a future PR.

web/dds_mvp_wasm.cpp
```diff
@@ -1,5 +1,5 @@
 /*
```
tzimnoch, line 1: This is your project and I haven't been able to find time to contribute meaningfully aside from some very minor changes and code review feedback. I appreciate if your response is that I'm free to implement what I want myself.

I think demonstrating how to call DDS's API directly gives other users maximum flexibility.

It's not customary to integrate with a third party library by patching its code and recompiling. It's certainly an option with FOSS. This file is currently part of the official release. A wasm artifact containing only the code from /library is not yet being provided.

I'd try to be mindful about what code going into here is really part of the dds library and what is part of the website you're building with the library. This could all be client side Javascript. I think more of it should be.

web/dds_mvp_wasm.cpp
```diff
@@ -16,13 +19,102 @@
...
+    *kind_is_small = (cfg.tt_kind_ == TTKind::Small) ? 1 : 0;
+  }
+  if (def_mb != nullptr) {
+    *def_mb = cfg.tt_mem_default_mb_;
+  }
+  if (max_mb != nullptr) {
+    *max_mb = cfg.tt_mem_maximum_mb_;
+  }
+}
+#endif
```
tzimnoch, line 116: Does this belong in dds_mvp_wasm_test.cpp? Is it possible to move there and still do the needful?

tameware:
Do you mean get rid of the C++ shim and call WASM entrypoints directly from
Javascript? I like that idea a lot. As you say, if it works it would be a
good example for others to follow. I’ll ask Cursor about it now and if it’s
straightforward I’ll have Cursor implement it in my next PR.

Please don’t worry that you’re under-contributing. I find your reviews
immensely valuable.

tameware:
Here’s what Cursor had to say:

https://cursor.com/agents/bc-019fc9ae-4c56-7a3a-96a4-84c15250c6f6

tameware:
Pasting this time:


You can’t drop every native export — JS can’t call
calc_dd_table_pbn(SolverContext&, …) directly. What you can remove is the
MVP-specific bridge in web/dds_mvp_wasm.cpp, if JS takes over marshalling
and context lifecycle against a real C ABI.


What the shim does today


dds_mvp_calc_table is the only browser entrypoint. It:


1. Validates / copies the PBN string into DdTableDealPBN

2. Owns a session-scoped static SolverContext and calls reset_for_solve()

3. Calls calc_dd_table_pbn

4. Flattens res_table[5][4] into int[20]


JS reaches it via Emscripten ccall (web/dds_mvp.js), with exports wired in
web/BUILD.bazel:



The library’s dds_c_api (dds_c_create_solvercontext_default,
dds_c_calc_dd_table, …) is for Java/FFM/.NET — it is not on the WASM export
list, and it has no PBN overload (binary DdTableDeal only).


Paths


| Goal | What’s involved |

| --- | --- |

| Call `_dds_mvp_calc_table` without `ccall` | Small JS change: write the
PBN into WASM memory yourself, call `module._dds_mvp_calc_table`\. C\+\+
bridge stays\. |

| Delete `dds_mvp_wasm.cpp` | Export a C surface \(extend `dds_c_api` with
something like `dds_c_calc_dd_table_pbn`, or use `CalcDDtablePBN` with
awkward by\-value structs\)\. Move PBN packing, `int[20]` flattening, and
context create/reset/destroy into JS\. Update `EXPORTED_FUNCTIONS`,
`dds_mvp.js`, Node smoke, e2e, and the native MVP unit tests\. |

| Drop Emscripten glue / raw `WebAssembly.instantiate` | Unrealistic while
using `-pthread` / `PTHREAD_POOL_SIZE` — you’d reimplement workers, memory
growth, and imports\.


Practical minimum to remove the MVP C++ file


1. Add a PBN-capable C export (or accept binary-deal marshalling + PBN
parse in JS).

2. Link that API into the WASM binary instead of dds_mvp_wasm.cpp.

3. Hold one DDS_C_SOLVER_CTX for the page lifetime (to keep TT reuse /
reset_for_solve behavior).

4. Replicate the current flat strain * 4 + hand layout (or change JS to
read DdTableResults in place).

5. Keep the Emscripten Module for _malloc / _free / heap access / pthreads.


Bottom line: some extern "C" WASM export remains. Removing the shim means
making the library C ABI (plus a PBN entry) the export surface and shifting
marshalling/lifecycle into JavaScript — not calling the C++ API “directly.”

tzimnoch:
I'm not sure if Cursor's write up clarifies or confuses things further for me.

Out of the box, Emscripten only works against extern "C" functions directly. If you want to call C++ functions then there will need to be an additional abstraction/renaming layer using embind https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html

In my project I'm working against DDS 2.9, so everything was C, hence using ccall and cwrap and exporting the C function names.

If this should commit to the C++ signatures, then maybe embind is the way to go, in which case at least one cpp file outside the base DDS /library code will be necessary.

It doesn't look like the build is currently taking advantage of emscripten's built-in javascript generation from the linker, i.e., the -o flag. https://emscripten.org/docs/tools_reference/emcc.html (see "-o \<target\>") This can do a lot of the heavy lifting.

I would defer this sort of change for now. It's outside the scope of this PR.

I don't know when I'll find copious free time, but I've done this before and can take a stab at it on this project eventually.

I still haven't switched my own project over to 3.0 because of the performance hit. I was hoping to invest more time in that first.

tameware:
I certainly did not get all of Cursor's explanation! If I understood correctly, we'd need another C entrypoint, which should not be a big deal. I have a pending PR that adds two entrypoints, and Martin has not approved it, so he might have concerns about such things.

I see no need to commit to the C++ signatures. That's just the method Cursor chose when I switched from calling out to a server to using Emscripten.

I agree the change is outside the scope of this PR. I'll ask Martin about the new entrypoint. If we add it first, that might help clarify things.

tameware:
> It doesn't look like the build is currently taking advantage of emscripten's built-in javascript generation from the linker, i.e., the -o flag. https://emscripten.org/docs/tools_reference/emcc.html (see "-o <target>") This can do a lot of the heavy lifting.

Per Cursor:

**Short answer: you're already using it** — switching further to “raw” `emcc -o` wouldn't buy much.

`wasm_cc_binary` for `//web:dds_mvp_wasm` already asks Emscripten for both glue and binary:

```33:41:web/BUILD.bazel
wasm_cc_binary(
    name = "dds_mvp_wasm",
    cc_target = ":dds_mvp_wasm_cc",
    threads = "emscripten",
    outputs = [
        "dds_mvp_wasm.js",
        "dds_mvp_wasm.wasm",
    ],
)
```

That’s the same outcome as `emcc … -o dds_mvp_wasm.js` (per the docs: `.js` → JS + separate `.wasm`). The modular flags (`MODULARIZE`, `EXPORT_NAME=createDdsModule`, exports) are also Emscripten link-time settings on that path.

What *looks* custom in `update_wasm.sh` is post-processing on top of that glue:

1. **`gen_wasm_bin_js.py`** — base64-embeds the `.wasm` so the page can pass `wasmBinary` without a separate fetch (useful for `file://` / simpler static serving).
2. **`patch_mvp_wasm.py`** — small `isFileURI` fix on the generated glue.

So `-o` isn’t the missing piece. The interesting lever, if you want less custom plumbing, would be something like **`-sSINGLE_FILE=1`**, which folds the wasm into the JS Emscripten emits. That could replace the base64 helper, but it conflicts with the current surgical patch strategy and doesn’t remove the need for COOP/COEP (pthreads) or the hand-written `dds_mvp.js` app layer.

**Bottom line:** the build already takes advantage of Emscripten’s JS generation from the linker. Extra `-o` usage wouldn’t help; simplifying the post-build embed/patch path (or adopting `SINGLE_FILE`) is the only related cleanup that might.

tameware:
@tzimnoch, please approve the PR when you have a chance.

tameware:
> I'll ask Martin about the new entrypoint.

Created Issue https://github.com/dds-bridge/dds/issues/284

tzimnoch (approved):

PR #291, opened 2026-08-06
Add par output to dd_table_for_deal

tameware:
## Summary
- Extend C++ and Python `dd_table_for_deal` to print compact par after the DD table (`--vul none|both|ns|ew`)
- Process every unique deal in multi-board PBN files (higher read limit, clear missing-file errors, dedupe exact deal strings)
- Match Python multi-level Par encodings (`45S`) to C++ (`4S +1`); extract C++ helpers into a tested library; add an anonymized multi-board example PBN

## Test plan
- [x] `bazelisk test //examples:dd_table_for_deal_test //python:dd_table_for_deal_par_test`
- [x] `bazelisk run //examples:dd_table_for_deal -- hands/example.pbn` and compare par with Python
- [x] `bazelisk run //python/examples:dd_table_for_deal -- --vul ns hands/multi_board.pbn` (unique deals only, blank lines between boards)
- [x] Confirm making contracts show `=` / `+N` and sacrifices show undertricks with declaring-side score

### DIFF
--- a/examples/dd_table_for_deal_lib.cpp
+++ b/examples/dd_table_for_deal_lib.cpp
@@ -14,6 +14,7 @@
 #include <cstdio>
 #include <regex>
 #include <string>
+#include <unordered_set>
 
 namespace dd_table_for_deal {
 namespace {
@@ -100,6 +101,21 @@ auto extract_deal_tags(std::string_view text) -> std::vector<std::string>
     begin = match[0].second;
   }
   return deals;
+}
+
+
+auto unique_deals(std::vector<std::string> const& deals)
+    -> std::vector<std::string>
+{
+  std::vector<std::string> unique;
+  std::unordered_set<std::string> seen;
+  unique.reserve(deals.size());
+  for (const auto& deal : deals)
+  {
+    if (seen.insert(deal).second)
+      unique.push_back(deal);
+  }
+  return unique;
 }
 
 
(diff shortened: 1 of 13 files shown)


tameware:
Addressed Copilot’s suppressed review notes in 19a7895:
- Documented numeric `--vul` codes (`0|1|2|3`) in C++ and Python usage/help
- Changed the stdin failure message to the generic “Cannot read PBN from stdin”

tameware:
I addressed Copilot's third suppressed comment – NS-view should in fact be declaring-side view. The other two are incorrect or pedantic – Copilot has no bridge knowledge.

tameware:
These were not intended to be examples, I just couldn't think of a better place for them. They're growing features. Maybe move them to new `utilities/` directories? If so, I'd do in a separate PR.

examples/dd_table_for_deal_lib.cpp
```diff
@@ -0,0 +1,290 @@
...
+#include <istream>
+#include <limits>
+#include <regex>
+#include <string>
+#include <unordered_set>
+
+namespace dd_table_for_deal {
+namespace {
+
+const std::regex DEAL_TAG_RE{
```
zzcgumn, line 27: This makes me wonder if we should add a pbn parser to the dds API.

tameware:
I’ve wondered the same!

tameware:
Ready to approve as is, @zzcgumn? Shall I add an "add a pbn parser to the dds API" issue?

zzcgumn (approved):

PR #296, opened 2026-08-07
Removed dtest --min/--max flags. 

tameware:
--min/--max reported minimum and maximum batch averages, not single-deal times, so they were useless.

Use -r for per-board outliers instead.

### DIFF
--- a/library/tests/args_test.cpp
+++ b/library/tests/args_test.cpp
@@ -213,8 +213,6 @@ TEST(Args, MakeDirFailsWhenPathIsExistingFile)
 
 TEST(Args, UnknownMinMaxFlagsAreRejected)
 {
-  // Former --min/--max batch-extreme flags must not be accepted.
-  // read_args prints to cout and exits(0); death tests only match stderr.
   const std::string path = make_temp_input_file();
   char arg0[] = "dtest";
   char arg_f[] = "-f";
--- a/library/tests/args_test.cpp
+++ b/library/tests/args_test.cpp
@@ -219,8 +219,11 @@ TEST(Args, UnknownMinMaxFlagsAreRejected)
   char arg0[] = "dtest";
   char arg_f[] = "-f";
   char arg_min[] = "--min";
-  char* argv[] = {arg0, arg_f, const_cast<char*>(path.c_str()), arg_min};
-  EXPECT_EXIT(read_args(4, argv), ::testing::ExitedWithCode(0), ".*");
+  char arg_max[] = "--max";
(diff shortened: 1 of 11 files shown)


tameware:
Addressed Copilot's note in fb39676: `UnknownMinMaxFlagsAreRejected` now asserts both `--min` and `--max` are rejected.

library/tests/args_test.cpp
```diff
@@ -222,42 +211,16 @@ TEST(Args, MakeDirFailsWhenPathIsExistingFile)
...
-  char arg0[] = "dtest";
-  char arg_f[] = "-f";
-  char arg_max[] = "--max";
-  char arg_min[] = "--min";
-  char* argv[] = {
-    arg0, arg_f, const_cast<char*>(path.c_str()), arg_max, arg_min};
-  read_args(5, argv);
-  EXPECT_TRUE(options.show_min_);
-  EXPECT_TRUE(options.show_max_);
+  EXPECT_EXIT(read_args(4, argv), ::testing::ExitedWithCode(0), ".*");
```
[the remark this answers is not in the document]
tameware: Addressed.

tameware:
Addressed Copilot's suppressed note in ff0023f: `ResetClearsAccumulatedStats` now requires the Number of hands field to be exactly `0` (not a substring match).

tameware:
> Grammar: “Non-positive hands is ignored” should be plural (“hands are ignored”).

No, it's correct as is.

> PR title/description focuses on removing dtest --min/--max, but this PR also changes Bazel test sizing (and removes a calc_par performance test). 

That's a cherry-pick of the orthogonal fix necessary to get the CI to pass. It will disappear in the merge.

library/tests/args_test.cpp
```diff
@@ -222,42 +211,19 @@ TEST(Args, MakeDirFailsWhenPathIsExistingFile)
...
-  char* argv[] = {arg0, arg_f, const_cast<char*>(path.c_str()), arg_max};
-  read_args(4, argv);
-  EXPECT_TRUE(options.show_max_);
-  EXPECT_FALSE(options.show_min_);
-}
-
-TEST(Args, MinFlagEnablesShowMin)
+TEST(Args, UnknownMinMaxFlagsAreRejected)
 {
+  // Former --min/--max batch-extreme flags must not be accepted.
```
zzcgumn: Please remove these comments about what the code used to do. I don't understand why coding agents believe they are useful.

zzcgumn (approved):
Makes sense.

library/tests/args_test.cpp
tameware: Removed.

PR #297, opened 2026-08-07
Avoid MSAN timeouts

tameware:
Drop the redundant ContextReusePerformance loop (reuse is already covered by ContextReuseMultipleCalls) and mark the target medium so the remaining suite has headroom under the MSAN small-test limit.


zzcgumn (approved):
Makes sense.

PR #309, opened 2026-08-11
Web: drag-and-drop cards and undeployed deck in diagram center

tameware:
## Summary
- Move undeployed cards into the hand-diagram center (four suit rows) and omit dealt cards from that strip instead of graying them out.
- Add HTML5 drag-and-drop between hands and the center: cards insert sorted high→low; full (13-card) hands reject drops; same-hand drags are a no-op.
- Align suit glyph columns / pip spacing across seats and center
- show hand card-count notes (singular “1 card”) whenever a hand is not empty and not exactly 13, left-aligned with the pips.
- Disallow entry of more than 13 cards per hand

Deployed to https://tameware.com/adam/bridge/dds/dds_web.html

## Test plan
- [ ] `bazel test //web:dds_web_js_test`
- [ ] `bazel test --test_tag_filters=e2e //web:dds_web_e2e_test`
- [ ] Manually drag cards among N/E/S/W and the center; confirm sort order and rejection when a hand already has 13
- [ ] Confirm card-count shows for partial hands (“1 card” / “3 cards”), hides at 0 and 13, and lines up with suit pips


web/dds_web.js
```diff
@@ -394,6 +406,14 @@ function allDeckCards() {
     return cards;
 }
 
+function undeployedCardHtml(card) {
+    return "<button type=\"button\" class=\"hand-card\" draggable=\"true\"" +
+        " data-card=\"" + escapeHtml(card.key()) + "\"" +
+        " tabindex=\"-1\" aria-hidden=\"true\">" +
+        escapeHtml(card.pip) +
+        "</button>";
+}
```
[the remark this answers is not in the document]
tameware: Fixed: removed `aria-hidden` from undeployed `.hand-card` buttons. Kept `tabindex="-1"` so they stay out of the tab order while remaining valid drag sources in the accessibility tree.

tameware:
Addressed the latest Copilot notes on this PR:

- **Center filler regex** (`dds_web_test.mjs`): updated to assert `.grid-filler-center` is immediately followed by `#deck-status` via opening tags, so nested `</div>` no longer truncates the match.
- **Caret / duplicate pips in `sanitizeHandSuitInputs`**: no change — within-suit duplicates are already stripped on input (`sanitizeSuitHolding` + beep), so `indexOf(pipBeforeCaret)` cannot hit a second occurrence in the sanitized holding.

tameware:
Addressed Copilot’s note on `handCardCount()`: it now sums the four suit input lengths for the target seat instead of calling `collectHands()` on every dragover.

tameware:
Addressed Copilot’s note on undeployed card accessible names: `undeployedCardHtml` now sets `aria-label` via `handCardAriaLabel(\"undeployed\", card)` (e.g. \"Undeployed heart ace\"), consistent with dealt hand-card buttons.

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+    assert.equal(document.element("north_spades").value, "AKQ");
+});
+
+test("addCardToHand inserts between existing ranks", () => {
+    const document = createMockDocument({ north_spades: "A" });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.addCardToHand("north", new ctx.Card("spades", "K")), true);
+    assert.equal(document.element("north_spades").value, "AK");
+});
```
tzimnoch, line 2531: These two tests are verifying the same thing, despite their descriptions. Both tests would pass if the addCardToHand simply added the card to the end of the suit without caring about order.

The first test should insert several cards out of order, add the 5, 2, Q, T, 7. Verify the result is "AKQT752".

tameware: Good catch — both cases were append-shaped. Updated:

- Out-of-order inserts `5, 2, Q, T, 7` onto `AK` → `AKQT752`
- Mid-suit insert `K` into `AQ` → `AKQ` (would be `AQK` if append-only)

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+        north_spades: "A",
+        east_hearts: "K",
+    });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.addCardToHand("south", new ctx.Card("spades", "A")), false);
+    assert.equal(document.element("south_spades").value, "");
+    assert.equal(ctx.addCardToHand("west", new ctx.Card("hearts", "K")), false);
+});
+
```
tzimnoch, line 2548: Test is asymmetrical. The check on line 2382 is not repeated for west.

Not crucial, but should there also be verification that north and east still hold the cards that were attempted to be assigned?

This is ultimately a UX issue. I find interfaces that require me to remove a card from one hand before giving it to another player to be clumsy. I don't want to have to remove a card from another player's hand before assigning it somewhere else. I just want it to move from where it was to where I want it to be.

I have performed no usability study to determine what a general user actually prefers.

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+test("removeCardFromHand deletes the first matching pip", () => {
+    const document = createMockDocument({ north_spades: "AKA" });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.removeCardFromHand("north", new ctx.Card("spades", "A")), true);
+    assert.equal(document.element("north_spades").value, "KA");
+    assert.equal(ctx.removeCardFromHand("north", new ctx.Card("spades", "Q")), false);
+});
```
tzimnoch, line 2579: Not sure you want this test case.  "AKA" should be rejected as a spade suit.

After removing a card, the remaining cards should be in rank order. "KA" shouldn't happen.

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+    assert.equal(document.element("east_hearts").value, "K");
+});
+
+test("moveCardToHand moves from undeployed into a hand in rank order", () => {
+    const document = createMockDocument({ north_spades: "AK" });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.moveCardToHand(new ctx.Card("spades", "Q"), "north"), true);
+    assert.equal(document.element("north_spades").value, "AKQ");
+});
```
tzimnoch, line 2599: These checks should be moderately harder. Set up should be AQ, then insert the K.

web/tests/dds_web_test.mjs
tameware: Agreed on the test asymmetry — updated so both rejected adds assert the destination stayed empty and the original owners (`north` / `east`) still hold the card.

On the UX point: drag-and-drop already auto-moves via `moveCardToHand` (remove from wherever it was, then place). The reject path here is for the lower-level `addCardToHand` / typed-input case, which currently refuses duplicates so typing an already-dealt card doesn’t silently steal it.

Happy to change typing to auto-transfer as well if you’d rather that be the default — want me to do that?

web/tests/dds_web_test.mjs
tameware: Agreed — `AKA` isn’t a valid holding (duplicates are stripped on input to `AK`), and a remove shouldn’t leave unsorted leftovers. Replaced the case with `AKQ` → remove `A` → `KQ`, and assert a second remove of `A` fails while leaving `KQ` unchanged.

web/tests/dds_web_test.mjs
tameware: Updated: start from `AQ`, insert `K` → `AKQ` (append-only would leave `AQK`).

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+    assert.equal(document.element("east_spades").value, "AK");
+});
+
+test("moveCardToHand rejects drag-and-drop within the same hand", () => {
+    const document = createMockDocument({ north_spades: "AKQ" });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.moveCardToHand(new ctx.Card("spades", "A"), "north"), false);
+    assert.equal(document.element("north_spades").value, "AKQ");
+});
```
tzimnoch: Not strictly necessary if the operation is idempotent. I don't know if you'll serve an error to the user when the function result is false, but if a player picks up a card, circles it around the screen a few times and returns it to the same hand they plucked it from, they shouldn't see an error.

As long as the hand doesn't accidentally get two Aces, it's fine for this to be a happy path.

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+        north_spades: "AKQJT98765432",
+        east_hearts: "A",
+    });
+    const ctx = loadDdsWeb(document);
+
+    assert.equal(ctx.moveCardToHand(new ctx.Card("hearts", "A"), "north"), false);
+    assert.equal(document.element("north_spades").value, "AKQJT98765432");
+    assert.equal(document.element("north_hearts").value, "");
+    assert.equal(document.element("east_hearts").value, "A");
+});
```
tzimnoch, line 2629: Again, this is a UX question.

Forcing the user to remove a spade first might cause frustration.

web/tests/dds_web_test.mjs
tameware: That was Copilot writing on my behalf. I never thought of having a typed card automatically remove it from another hand. I like the idea. It's orthogonal to this PR - I'll add it to my ToDo list.

web/tests/dds_web_test.mjs
tameware: I recently made a design decision that no hand should allow more than 13 cards, but without much conviction. I can revisit it. I think it's OK for now.

web/tests/dds_web_test.mjs
tameware: I confirmed empirically - no error message, and I see none in the test. Should I add a "Does not display error message" assertion?

web/tests/dds_web_test.mjs
tameware: Changed my mind - I'll add moving typed cards to this PR.

tameware:
> * Please either (a) implement a hard cap for typed/pasted input (reject/trim additional cards beyond 13 for the edited hand)

Done.

web/tests/dds_web_test.mjs
```diff
@@ -2149,6 +2272,516 @@ test("dds_web html does not cache-bust css or js with query params", () => {
...
+    const ctx = loadDdsWeb(createMockDocument());
+    const card = new ctx.Card("spades", "\"&<>'");
+
+    const html = ctx.undeployedCardHtml(card);
+
+    assert.match(
+        html,
+        /aria-label="Undeployed spade &quot;&amp;&lt;&gt;&#39;"/
+    );
+});
```
tzimnoch: I don't see a use case for this.

new ctx.Card("inͭͬv̯͐a|id̛̑", "in͕̠p̯\/t") should probably throw an exception.

tameware:
Addressed the remaining review threads:

- **Same-hand drag/drop**: dropping a card back onto its own hand is now an idempotent no-op (`moveCardToHand` returns true, hand unchanged) instead of a rejected drop.
- **Invalid `Card` values**: `Card` / `Card.fromKey` now throw on invalid suit, pip, or key; removed the redundant undeployed aria-label escape test that constructed garbage cards. `collectHands` skips non-bridge pips before constructing cards.

tameware:
Addressed Copilot’s follow-up: dragover/drop now use `cardFromKeySafe` so invalid payload keys are treated as non-drops instead of throwing.

tameware:
Addressed Copilot’s note: `cardFromKeySafe` now uses `catch (_err)` to match surrounding style.

tameware:
Addressed Copilot’s note on `Card.fromKey`: keys must be exactly two characters and a string; case is normalized before suit/pip parsing so malformed drag payloads are rejected cleanly.

tameware:
Addressed Copilot’s a11y note: `#deck-status` now has `role="group"` so its `aria-label` is reliably exposed to assistive tech.

tameware:
Good to go, @tzimnoch ?

web/tests/dds_web_test.mjs
```diff
@@ -1040,44 +1228,86 @@ test("updateActionButtons grays cards entered in any hand, including lowercase p
...
+    });
+    const ctx = loadDdsWeb(document);
+
+    ctx.updateActionButtons();
+
+    assert.equal(document.element("north_spades").value, "AKQJT98765432");
+    assert.equal(document.element("north_hearts").value, "");
+    assert.equal(document.element("north-card-count").hidden, true);
+    assert.equal(document.element("north-card-count").innerHTML, "");
+});
```
tzimnoch: I find things like this awkward because you cannot know which card the user would have wanted to remove from the hand to get it back down to 13 cards.

tzimnoch (approved):

web/tests/dds_web_test.mjs
tameware: Good call. I've removed this test and the code it was testing. I'll merge after one more Copilot review.

PR #310, opened 2026-08-11
Fix for WASM on iOS WebKit

tameware:
## Summary
- Force `coi-serviceworker` to use `COEP: require-corp` instead of `credentialless`, which Safari / iOS WebKit (including Chrome and Firefox on iPhone) do not treat as cross-origin isolation.
- DDS Web assets are all same-origin, so `require-corp` is sufficient and makes `SharedArrayBuffer` / pthread WASM reliable on GitHub Pages for iPhone browsers.
- Add an HTML unit test that the `window.coi` config is set before the worker script loads.

## Test plan
- [x] `bazel test //web:dds_web_html_test` (or `python3 -m unittest web.tests.test_web_html`)
- [ ] After merge + Pages deploy, open https://dds-bridge.github.io/dds/ on iPhone Safari/Firefox/Chrome
- [ ] Clear site data once if an old COI worker is stuck, reload, enter a test deal, confirm DD table solves (no SharedArrayBuffer error)
- [ ] Spot-check desktop Chrome/Firefox still solve as before

### DIFF
--- a/web/dds_web.html
+++ b/web/dds_web.html
@@ -17,6 +17,11 @@
 <html lang="en">
 <head>
     <meta charset="utf-8">
+    <!-- Safari / iOS WebKit (incl. Firefox on iPhone) ignore COEP credentialless
+         for isolation. DDS assets are same-origin, so require-corp is enough. -->
+    <script>
+    window.coi = { coepCredentialless: () => false };
+    </script>
     <script src="coi-serviceworker.js"></script>
     <link rel="stylesheet" type="text/css" href="dds_web.css">
     <title>
--- a/web/tests/test_web_html.py
+++ b/web/tests/test_web_html.py
@@ -58,6 +58,25 @@ class DdsWebHtmlCoiTest(unittest.TestCase):
         self.assertLess(coi_at, wasm_at)
         self.assertLess(coi_at, app_at)
 
+    def test_disables_coep_credentialless_before_coi_serviceworker(self) -> None:
+        # coi-serviceworker defaults to COEP: credentialless. Safari / iOS WebKit
+        # (including Firefox on iPhone) do not honor that value for isolation, so
+        # SharedArrayBuffer stays unavailable. DDS Web is same-origin only, so
+        # require-corp is correct — configure window.coi before the worker script.
+        text = HTML_PATH.read_text(encoding="utf-8")
+        head_match = re.search(r"<head\b[^>]*>(.*?)</head>", text, re.I | re.S)
+        self.assertIsNotNone(head_match)
+        head = head_match.group(1)
+        config_at = head.find("coepCredentialless")
+        worker_at = head.find('src="coi-serviceworker.js"')
+        self.assertNotEqual(config_at, -1, msg="must set window.coi.coepCredentialless")
+        self.assertNotEqual(worker_at, -1)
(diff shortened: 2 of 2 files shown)


web/dds_web.html
```diff
@@ -17,6 +17,11 @@
 <html lang="en">
 <head>
     <meta charset="utf-8">
+    <!-- Safari / iOS WebKit (incl. Firefox on iPhone) ignore COEP credentialless
+         for isolation. DDS assets are same-origin, so require-corp is enough. -->
+    <script>
+    window.coi = { coepCredentialless: () => false };
+    </script>
     <script src="coi-serviceworker.js"></script>
```
tzimnoch, line 25: I somehow do not have this file in my local copy of develop. I just fetched and pulled and it's still not there.

I may have to reclone?

tameware: I'm far from expert on Git. I think this is expected. To pull it in you could try `gh pr checkout 310`

web/dds_web.html
```diff
@@ -17,6 +17,11 @@
 <html lang="en">
 <head>
     <meta charset="utf-8">
+    <!-- Safari / iOS WebKit (incl. Firefox on iPhone) ignore COEP credentialless
+         for isolation. DDS assets are same-origin, so require-corp is enough. -->
+    <script>
+    window.coi = { coepCredentialless: () => false };
+    </script>
```
tzimnoch, line 24: I don't think I'm qualified to weigh in on this one. I would expect to make the adherence to cross origins rules more strict rather than relaxing them.

I don't understand how this would fix something on iOS that the comment says is ignored there.

I'm not the right person to approve this.

tameware: If I understand correctly, it's making the rule stricter. It's a double negative: Credentialless false means require credentials. Per my AI:

> Cross-Origin-Embedder-Policy: require-corp (COEP) tells the browser that a document may load cross-origin subresources only when those resources explicitly permit it. 

We don't load anything but our own code, so we don't need `Credentialless`.

tameware: That said, I'll merge this tomorrow even with no approval so that we can continue testing the Android issue that Klaus Krtschil sent email about. It's a small PR, easy to roll back if it breaks something.

krtschil: After a "git pull" I don't get a revised dds_web.html. Git said that my clone is up-to-date. Also the file in the develop branch still does not contain your change.
I can't review this as I do not have an iOS device for testing.

tameware: Per my reply to @tzimnoch above, I think this is expected. `gh pr checkout 310` should bring it in. 

Deploying the PR to github pages without merging is tricky. I'll merge tomorrow (Aug 12) and report back on the email thread. Were I more diligent, I'd move the thread to a GitHib Issue.

tzimnoch: I somehow missed a word on my comment above.

"I would expect -this- to make the adherence to cross origins rules more strict"

I agree that it's going to make the rules stricter. I don't understand how that's going to fix an issue.

Sometimes the proof is in the pudding. Roll it out and verify.

tameware:
It seems to have worked. No idea why.

I suspect it will keep me from implementing things like direct Handviewer
URL reading, as you have on your site.

PR #314, opened 2026-08-13
Add create_list_for_dtest for generating dtest hand-list files

tameware:
## Summary
- Add `//python/utilities:create_list_for_dtest` (`python/utilities/src/create_list_for_dtest.py`) to generate seeded random dtest hand-list files with FUT/TABLE/PAR/PAR2/PLAY/TRACE filled from DDS.
- Add `utilities/src/regenerate_hand_lists.sh` and `utilities/src/verify_lists.sh` to batch-generate and verify lists.
- Stream deals one at a time to the output file, validate deal counts against dtest limits, reuse a `SolverContext` across solves, and cover helpers plus an end-to-end `main()` path in unit tests.
- Clarify dtest README hand-list format notes and the seed/reproducibility docstring.

## Test plan
- [x] `bazelisk test //python/utilities:create_list_for_dtest_test`
- [x] `bazelisk run //python/utilities:create_list_for_dtest -- -n 1 --seed 1 -o /tmp/list1.txt` and inspect output tags
- [x] Confirm invalid `-n 0` / `-n 100001` exit with argparse errors


tameware:
> * `create_list_for_dtest_test` is declared as a Bazel `size = "small"`, but the test suite runs many DDS solves (e.g., `generate_dd_play` performs 52 `solve_board_pbn` calls per generated deal, plus `analyse_play_pbn` and table/par computations). This is closer to an integration/performance-style test and is more likely to exceed small-test time limits under CI load variance (similar to `library/tests/solve_board:analyse_play_consistency_test` being marked `medium`). Consider bumping the size to `medium`, or alternatively mock DDS calls in the end-to-end `main()` test to keep this target truly small.

Bazel’s default timeout for small is 60s. This suite ran 34 tests in 0.158s, so it’s nowhere near that budget.

tameware:
Addressed the latest suppressed Copilot notes:

- `utilities/src/verify_lists.sh`: header now says failure text is checked in the combined stdout/stderr capture.
- `utilities/src/regenerate_hand_lists.sh`: header now says output goes to `OUT_DIR` (default under `/tmp`), not in-place under `hands/`.
- `create_list_for_dtest`: `main()` reuses one `SolverContext` through `solve_fut` / `generate_dd_play` for large `-n` runs.
- PR description updated to `//python/utilities:create_list_for_dtest` and the utilities scripts paths.

zzcgumn (approved):
Feels like a good addition. We want to be able to generate more regression examples.

PR #316, opened 2026-08-14
Enable benchmark.py on Windows

tameware:
## Summary
- Remove the `target_compatible_with` guard that skipped `//python:benchmark_test` on Windows.
- Add `ensure_executable()` so branch-binary copies skip Unix `chmod` on Windows.
- Add unit tests for the cross-platform executable helper.

## Test plan
- [x] `bazel test //python:benchmark_test`
- [x] Windows CI `bazel test //...` includes `//python:benchmark_test`


zzcgumn (approved):
It is a very irritating fact that we have to do things like this to support all major platforms.

PR #323, opened 2026-08-15
Stabilize dd-table-before-leads web JS ordering test

tameware:
## Summary
- Fixes a flaky `//web:dds_web_js_test` failure where a late `dd-end` from `fillForm` could land after clearing the recorded event order.
- Drains the auto-solve queue and awaits the lead-refresh promise instead of fixed timeouts.

Split out of #315 (hand-list work); unrelated to that PR.

## Test plan
- [ ] `bazelisk test //web:dds_web_js_test --nocache_test_results` (repeat a few times)
- [ ] CI – Linux `build_and_test` green

### DIFF
--- a/web/tests/dds_web_test.mjs
+++ b/web/tests/dds_web_test.mjs
@@ -42,6 +42,15 @@ function findDdsWebJsPath() {
     throw new Error("dds_web.js not found");
 }
 
+/** Reject if `promise` does not settle within `ms` (clears the timer either way). */
+function withTimeout(promise, ms, message) {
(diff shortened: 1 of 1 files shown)


tameware:
Addressed the Copilot note: `leadsDone` now rejects if `refreshOpeningLeadTricks` throws, and is awaited via `withTimeout(..., 1000, ...)` so a missing lead refresh fails fast with a clear message instead of hanging until the suite timeout.

zzcgumn (approved):

PR #343, opened 2026-08-20
Enable Bazelisk Caching.

zzcgumn:
Caching downloads and some artefacts is expected to speed up our CI pipeline.

### DIFF
--- a/.github/actions/setup-bazelisk/action.yml
+++ b/.github/actions/setup-bazelisk/action.yml
@@ -8,12 +8,21 @@ inputs:
 runs:
   using: composite
   steps:
+    # disk-cache holds compiled action outputs, which are sensitive to each
+    # job's own flags (--define=ab_stats, --config=asan, compilation_mode,
+    # etc.), so it is scoped per workflow+job to avoid one job restoring
+    # another's incompatible outputs. repository-cache holds downloaded
+    # external repos (emsdk, LLVM, ...), which are the same for every job
+    # driven by the same MODULE.bazel/WORKSPACE, so it is left unscoped and
+    # shared.
     - name: Setup Bazelisk (attempt 1)
       id: setup_bazelisk_1
       continue-on-error: true
       uses: bazel-contrib/setup-bazel@0.19.0
       with:
         bazelisk-version: ${{ inputs.bazelisk-version }}
+        disk-cache: ${{ github.workflow }}-${{ github.job }}
+        repository-cache: true
 
     - name: Setup Bazelisk (attempt 2)
       id: setup_bazelisk_2
@@ -22,6 +31,8 @@ runs:
       uses: bazel-contrib/setup-bazel@0.19.0
       with:
         bazelisk-version: ${{ inputs.bazelisk-version }}
+        disk-cache: ${{ github.workflow }}-${{ github.job }}
+        repository-cache: true
 
     # Composite run steps need an explicit shell. Use bash on Linux/macOS (pwsh
     # is not always present after free-disk-space) and pwsh on Windows.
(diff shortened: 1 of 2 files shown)


tameware:
I see the latest Windows run was 17 minutes. Is that faster than before the change? Do we have access to a log of build duration over time?

zzcgumn:
Claude checked build times so there must be a log somewhere, but I need to find out where it is.

tameware:
> I see the latest Windows run was 17 minutes. Is that faster than before the change? Do we have access to a log of build duration over time?

My mistake! It seems so far the caching applies only to the Linux build. We should look into that one too, but my concern is always with the one that's gating throughput.

tameware:
> > I see the latest Windows run was 17 minutes. Is that faster than before the change? Do we have access to a log of build duration over time?
> 
> My mistake! It seems so far the caching applies only to the Linux build. We should look into that one too, but my concern is always with the one that's gating throughput.

Or at any rate, `ci_linux.yml‎` is the only yml file changed…

zzcgumn:
> Or at any rate, ci_linux.yml‎ is the only yml file changed…

Yes, but is only changing the order of the steps to have those that similar things close to each other in the script. It was confusing to read. Speedup is expected for caching.

zzcgumn:
> my concern is always with the one that's gating throughput.

Linux build_and_test was 20 minutes for the release branch. 14 minutes here. Maybe it did benefit from the caching?

tameware:
> > my concern is always with the one that's gating throughput.
> 
> Linux build_and_test was 20 minutes for the release branch. 14 minutes here. Maybe it did benefit from the caching?

I hope so! Whenever I looked, Windows was the only one pushing 20 minutes. 

Cursor was helpful – the gh commands work nicely:

> For **dds-bridge/dds**, previous CI durations are on the Actions runs themselves.
> 
> ### In the GitHub UI
> 1. Open [Actions](https://github.com/dds-bridge/dds/actions)
> 2. Pick a workflow (e.g. `ci_linux`, `ci_macos`, `ci_windows`)
> 3. Open a past run — the header shows total duration
> 4. Expand jobs/steps for per-job and per-step times
> 
> ### With `gh` (CLI)
> List recent runs with durations:
> 
> ```bash
> gh run list --limit 20
> ```
> 
> Filter by workflow:
> 
> ```bash
> gh run list --workflow=ci_linux.yml --limit 20
> ```
> 
> Inspect one run (includes timing details):
> 
> ```bash
> gh run view <run-id>
> ```
> 
> Job timing only:
> 
> ```bash
> gh run view <run-id> --json jobs --jq '.jobs[] | {name, startedAt, completedAt}'
> ```
> 
> ### Trends over many runs
> GitHub doesn’t chart historical build times by default. Options:
> - Export via `gh run list` / the [Actions API](https://docs.github.com/en/rest/actions/workflow-runs) and plot yourself
> - Use a badge/service (e.g. workflow status badges) or an Actions analytics tool if you want ongoing tracking
> 
> If you say which workflow you care about (Linux/macOS/Windows/WASM), I can pull recent durations for it.

tameware:
```
% gh run list --limit 100 | grep 'CI – Linux' | awk -F'\t' '{print $8}' | sort
10m38s
12m26s
12m46s
12m53s
13m0s
13m33s
13m39s
13m59s
14m16s
14m21s
14m44s
14m8s
14m9s
15m51s
20m19s
20m21s
```

zzcgumn:
> I see the latest Windows run was 17 minutes.

I believe the Windows run will benefit from the caching. I think the best way forward is to merge this and check CI run times once we have a bit more data. Do you agree @tameware?

tameware:
I do. Easy to roll back if need be.

tameware (approved):

PR #349, opened 2026-08-24
Validate par and table inputs; add fuzz harnesses and SECURITY.md

zzcgumn:
## Summary

Closes five memory-safety defects in the input-validation and error-reporting
paths, and adds the fuzzing that found four of them.

The starting point was an audit question about whether DDS's input handling
warranted more attention. It did: `SolveBoard()` validates its input
thoroughly, but the par and `CalcDDtable*` entry points did not, and two of the
crashes turned out to be *inside* the validation and error-reporting logic
itself. Every fix is a range check; none touches the search algorithms.

**Fixes**

- **Par table.** `SidesParBin()` never checked `res_table`, so an out-of-range
  table overflowed `char temp[8]` in `Par()` (ASan: stack-buffer-overflow) and
  silently wrote 26 characters into a `char[10]` via `SidesPar()` — both while
  returning `RETURN_NO_FAULT`. Adds `par_table_checks()` and
  `RETURN_PAR_TABLE_FAULT` (-401), mirrored in `DdsStatus.java` and
  `ErrorMessage()`.
- **`DealerPar()` parameters.** Neither `vulnerable` (indexes `VUL_LOOKUP[4][2]`)
  nor `dealer` was validated; a negative `dealer` reached
  `NUMBER_TO_PLAYER[static_cast<unsigned>(pno)]`, where the cast turns -1 into
  4294967295. Both are range-checked, and the unsigned casts are replaced with
  guarded helpers.
- **`CalcDDtable*` card counts.** No check that the four hands held equal
  numbers of cards, so a 51-card deal reached the search and read 14248 bytes
  past `rel_rank_storage`. Reachable from an ordinary PBN file short of a card —
  the reproducer contains only legal PBN characters. Adds
  `table_deal_checks()`, enforcing exactly the rules `SolveBoard()` already
  applied, so nothing is rejected that the solver would have accepted.
- **`DumpInput()`.** Indexed `card_suit[5]`, `card_hand[4]` and `card_rank[16]`
  with the very values `board_range_checks()` was rejecting as out of range.
  Present in release builds, since it is compiled in unless
  `DDS_NO_DUMP_ON_ERROR` is defined.
- **`board_value_checks()`.** Validates `currentTrickSuit[k]` only when the
  matching rank is non-zero, but `hand_rel_first` derives from the card count
  rather than the trick entries, so a five-card deal with zero ranks indexed
  `remainCards` with an unchecked suit.

**Fuzzing**

Sanitizers already ran in CI, but nothing generated inputs for them. Adds
libFuzzer harnesses for `convert_from_pbn()`, `CalcDDtablePBN()`,
`SolveBoard()` and the par entry points. Each is also an ordinary `cc_test`
that replays a checked-in corpus, so it needs no libFuzzer and runs under
`--config=asan/ubsan` in normal CI — that is what keeps these fixed. The new
`--config=fuzz` uses the hermetic LLVM toolchain, since Apple's ships no
libFuzzer.

**SECURITY.md**

States the input trust model explicitly: DDS is an in-process library that
assumes trusted, well-formed input, with per-entry-point detail on what is and
is not validated, guidance for anyone exposing it to untrusted data, and
reporting via GitHub private advisories.

## Notes for reviewers

- No ABI change. `RETURN_PAR_TABLE_FAULT` is additive; struct layouts are
  untouched.
- The `CalcDDtable*` validation is deliberately no stricter than
  `SolveBoard()`'s. It rejects unbalanced, duplicate-card and out-of-range-bit
  deals — inputs that previously produced a crash or meaningless table.
- One open hardening item is left deliberately: `convert_from_pbn()` still
  silently skips unrecognised characters, so an invalid rank now yields
  `RETURN_CARD_COUNT` rather than `RETURN_PBN_FAULT`. Tightening it risks
  rejecting PBN files with trailing newlines, so it is documented rather than
  changed. See `library/tests/fuzz/findings/README.md`.
- `SECURITY.md` points reporters at GitHub private advisories because the repo
  had no security contact; maintainers may prefer a specific address.

## Test plan

- [x] `bazel test //library/...` — 55/55
- [x] `bazel test --config=asan //library/...` — 55/55
- [x] `bazel test --config=ubsan //library/...` — 55/55
- [x] `bazel test //python/...` (19) and `//jni/...` (5)
- [x] Each fix verified by reverting it with the tests in place: the suite
      reproduces the original ASan report, and passes with the fix restored
- [x] Post-fix campaigns: ~1.35M executions across the four harnesses
      (par 500k, pbn 500k, solve_board 300k, calc_dd_table_pbn 52k/240s) with
      no crashes
- [x] CI on Linux and Windows/MSVC (only macOS was available locally)

[LLVM Fuzz Tooling](https://llvm.org/docs/FuzzingLLVM.html)


tameware:
I'll have a look once all comments are resolved and we have a clean Copilot review. I still have Copilot credits - let me know if you'd like a run.

zzcgumn:
@tameware , could you give this another Copilot review please?

tameware:
In progress.

tameware:
Copilot finished, but did not report in the thread:

"Verdict: Ready to merge. All fixes are correct, well-tested with fuzz corpus replays under ASan/UBSan, and the PR description accurately represents the changes."

I'll have a quick look later tonight.

zzcgumn:
I have heard about fuzz testing before but it was news to me that LLVM has a ready made library available for it.

doc/dll-description.html
```diff
@@ -1120,14 +1120,14 @@ <h2 id="functions">Functions</h2>
...
-<p>The user can suggest to DDS a number of threads by calling SetMaxThreads. DDS will never create more threads than requested, but it may create fewer if there is not enough memory, calculated as above. Calling SetMaxThreads is optional, not mandatory. DDS will always select a suitable number of threads on its own.</p>
-<p>It may be possible, especially on non-Windows systems, to call SetMaxThreads() actively, even though the user does not want to influence the default values. In this case, use a 0 argument.</p>
-<p>SetMaxThreads can be called multiple times even within the same session. So it is theoretically possible to change the number of threads dynamically. </p>
+<p>SetMaxThreads no longer influences the thread count: its argument is ignored and internal batch threading was removed. To cap workers, pass an explicit maxThreads to the *N or *X entry points, or manage concurrency in the calling application (typically one SolverContext per thread). DDS will otherwise select a suitable number of threads on its own.</p>
```
tameware: We added back some internal threading.

library/src/calc_tables.cpp
```diff
@@ -317,6 +322,17 @@ int STDCALL CalcAllTablesN(
...
 
+  // dealsp->deals is a fixed MAXNOOFTABLES * DDS_STRAINS array, and the
+  // capacity check below multiplies by count. Bound no_of_tables first, so
+  // that multiply cannot overflow signed int and wrap past the check, and so
+  // the per-deal loops cannot read past the array.
+  if (dealsp == nullptr)
+    return RETURN_UNKNOWN_FAULT;
+
+  if (dealsp->no_of_tables < 0 ||
+      dealsp->no_of_tables > MAXNOOFTABLES * DDS_STRAINS)
```
tameware, line 333: Could be clearer. Maybe rename to MAX_NO_OF_TABLES_PER_STRAIN and also add a new constant for the product. There are a few occurrences.

tameware (approved):
Great stuff!

library/src/calc_tables.cpp
zzcgumn: Agree, but this not a new constant so I will update this is in a separate pull request. We should also update from `#define MAXNOOFTABLES  40` to use a `constexpr`. We may also want to make this a default value and let callers override it.

PR #362, opened 2026-09-01
Solution Update

WindowsNT:
1. Static Builds
2. All cpps in place
3. /MP added


zzcgumn:
@WindowsNT, CI should pass if you update your branch with the latest commits to develop.

PR #363, opened 2026-09-01
fix: restore dll.h include broken by header-split / deprecate-c-api merge

### LINKED ISSUE #359: Upstream regression
After the latest updates to library/src I tried to update my BSOL solution for the wasm but failed with a couple of similar errors:

`api/dds_c_api.cpp:261:16: error: use of undeclared identifier 'Par'`

After analysing the changes my digital assistent told me that there is an upstream regression:

```
/* Upstream regression fix: dds_c_api.h used to pull in <api/dll.h>
   transitively (which declares the legacy PascalCase functions this
   file forwards to directly: Par, SidesPar, DealerPar, DealerParBin,
   SidesParBin, ConvertToDealerTextFormat, ConvertToSidesTextFormat,
   GetDDSInfo, ErrorMessage). A later refactor replaced that with the
   narrower <api/dds_c_data_types.h> (struct types only, no function
   declarations), but this file was never updated to include dll.h
   directly - so those calls now fail with "use of undeclared
   identifier" against a fresh clone of the develop branch. This file's
   own top-of-file comment says it forwards "straight to the
   corresponding legacy dll.h function", confirming the intent; this
   just restores the include that intent already assumed was present. */
#include <api/dll.h>
```
The fix to library/src/api/dds_c_api.cpp solves the issue.

### LINKED ISSUE #360: **Build broken after #354/#355**  `bazelisk build //...` fails with undeclared identifiers in `dds_c_api.cpp`:
library/src/api/dds_c_api.cpp:261:16: error: use of undeclared identifier 'Par'
library/src/api/dds_c_api.cpp:275:16: error: use of undeclared identifier 'SidesPar'

Reproducible on clean develop (437e06da) with no local changes.

zzcgumn:
## Summary
- PR #355 (`reorganise_headers`) replaced the transitive `<api/dll.h>` include in `dds_c_api.h` with `<api/dds_c_data_types.h>`; PR #354 (`deprecate_old_c_api`) added context-free wrappers that call legacy `dll.h` functions (`Par`, `SidesPar`, `DealerPar`, `GetDDSInfo`, `ErrorMessage`, ...).
- The two merged without a textual conflict, leaving `dds_c_api.cpp` with no declarations for those functions and `dds_c_api_test.cpp` with incomplete legacy struct types (`DealPBN`, `ParResultsDealer`, `ParResultsMaster`, `ParTextResults`, `DDSInfo`).
- Fix: `#include <api/dll.h>` directly in the two translation units that use the legacy symbols, rather than re-adding it to the now C-consumable public header.

## Test plan
- [x] `bazel build //...`
- [x] `bazel test //library/tests:dds_c_api_test`

Closes #360 
Closes #359

### DIFF
--- a/library/src/api/dds_c_api.cpp
+++ b/library/src/api/dds_c_api.cpp
@@ -18,6 +18,7 @@
 
 #include <api/dds_c_api.h>
 #include <api/dds_api.hpp>
+#include <api/dll.h>   /* legacy Par/SidesPar/DealerPar/.../GetDDSInfo/ErrorMessage */
 
 /* This shim is the boundary between the C++ implementation and pure-C FFI
    consumers (JVM/FFM, .NET, ctypes). Two things must never cross it:
--- a/library/tests/dds_c_api_test.cpp
+++ b/library/tests/dds_c_api_test.cpp
@@ -20,6 +20,7 @@
 #include <string>
 
 #include <api/dds_c_api.h>
+#include <api/dll.h>  // full definitions for DealPBN, ParResults{Dealer,Master}, ParTextResults, DDSInfo
 
 namespace {
 


krtschil (approved):
Looks good

PR #364, opened 2026-09-02
Remove `using namespace std` / using-declarations from headers

### LINKED ISSUE #357: Remove `using std` from header files.
This is a poor habit that we inherited from the version 2.9.

zzcgumn:
## What

`using namespace std;` (and one namespace-scope `using std::chrono::time_point;`) appeared at file scope in several headers. That injects names into every translation unit that includes the header, directly or transitively — bad style and a maintenance hazard.

### Headers cleaned (declarations now fully qualified)

- `library/src/system/system.hpp`
- `library/src/system/thread_mgr.hpp`
- `library/src/system/time_stat.hpp`
- `library/src/system/timer.hpp`
- `library/tests/TestTimer.hpp`

The `using Clock = std::chrono::steady_clock;` alias-declarations are left as-is (they are type aliases, not using-directives/declarations).

### Translation units updated

Several `.cpp` files had been leaning on those headers' transitive `using`. They now declare what they need locally:

- Specific `using std::...;` declarations added to `init.cpp` and the `system/*.cpp` printing/formatting code.
- Point qualification (`std::vector`, `std::chrono::time_point`) in `calc_tables.cpp`, `solve_board.cpp`, `timer.cpp`, `TestTimer.cpp`.
- `testcommon.cpp` gains `using std::left; using std::right;` alongside its existing `using std::` block.

### Coding standard

`.github/instructions/cpp.instructions.md` gains a dedicated **Namespace `using`** section replacing the terse "No `using namespace` in headers" bullet: headers ban namespace-scope using-directives *and* using-declarations (class/function-body `using` and alias-declarations still fine); `.cpp` files prefer specific using-declarations, with file-scope `using namespace` permitted but discouraged; never rely on a `using` supplied by an included header.

## Testing

- `bazel build //...` — passes
- `bazel test //...` — 93/93 pass

Closes #357

### DIFF
--- a/.github/instructions/cpp.instructions.md
+++ b/.github/instructions/cpp.instructions.md
@@ -140,13 +140,32 @@ auto result = compute_area(
 - **RAII** – manage resources via destructors.
 - **Smart pointers** – prefer `std::unique_ptr` / `std::shared_ptr`.
 - **Const-correctness** – mark data/functions const where possible.
-- **No `using namespace` in headers.**
+- **Namespace `using`** – see [Namespace `using`](#namespace-using) below.
 - **C++ Core Guidelines** – [Style Guide](https://isocpp.github.io/CppCoreGuidelines/) follow naming, formatting, and layout.
 - **Enums** – use `enum class`.
 - **Functions** – keep short; extract helpers if > 20 lines.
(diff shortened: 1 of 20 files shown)


krtschil (approved):
Good to go

PR #373, opened 2026-09-13
Extract store_ab_tt_result from ab_search_0 

tameware:
So TT stores can be tested without changing the search path.


zzcgumn (approved):
Good spot, better test coverage always helps. I am guessing this makes it easier to work on improving the transposition table and  lookups.

PR #376, opened 2026-09-14
Add a parallel Linux CI job that runs dtest on list100

### LINKED ISSUE #334: dtest enhancements
1. Should return non-zero exit status when it encounters errors. That way it can more easily be used in regression tests.
2. Should run in the CI.
3. Should use ANSI escape sequences when reporting progress, hide at the end of the run.
4. Should report progress when handling -s calc, including GIB-style solXXX.txt files.
5. Should test all five strains. Currently NT is omitted.

tameware:
## Summary
- Add a dedicated `dtest` job on Linux CI that runs `bazelisk run //library/tests:dtest -- -f hands/list100.txt` in parallel with `build_and_test` and the sanitizer jobs.

## Test plan
- [x] Confirm the new `dtest` job appears on the PR checks and passes
- [x] Confirm it does not block or serialize behind `build_and_test`


tameware:
I can add a dtest against masterDD.txt, with 80K+ deals. The new run will still finish before the others. What do you think?

zzcgumn:
I would add the dtest against masterDD.txt as a separate test that this triggered manually first. We can then decide whether to include it in the standard CI run based on the runtimes we see.

My view is that we merge these changes as is and add the dtest against larger file in a separate pull request.

zzcgumn (approved):

PR #380, opened 2026-09-14
Replace test-deal buttons with a Sample deals dropdown

tameware:
## Summary
- Replace the three test-deal toolbar buttons with a single **Sample deals…** dropdown (NS make 7, Everyone makes 3N, Part-score).
- Style toolbar buttons and the dropdown with the same flat 1px chrome; the select keeps a chevron as its only visual difference.
- Update HTML/JS unit tests and e2e helpers for the new control.

## Test plan
- [ ] Load DDS Web and confirm Clear / Import / Sample deals / Rotate share the same flat button look
- [ ] Choose each Sample deals option and confirm the corresponding deal loads, then the placeholder returns
- [ ] Re-select the same sample after editing and confirm it loads again
- [ ] `python3 -m unittest web.tests.test_web_html`
- [ ] `node --test web/tests/dds_web_test.mjs`


tameware:
Addressed the Copilot coverage note: added unit and e2e coverage that loads a sample deal, edits a hand, then selects the same Sample deals option again (depends on resetting the select to the placeholder).

tzimnoch (approved):

Issue #382, opened 2026-09-15
Refactor dds_web.js


tameware:
Per @[tzimnoch](https://github.com/tzimnoch)

This file is getting long -- almost 3000 lines. The pbn, lin, dlm import functions are prime candidates for existing in their own file(s) and being imported. It will also make clearer which of the functions above solely support file loading and are not used by the UI or DDS solver.

PR #385, opened 2026-09-15
Refactor dds_web.js into import, core, solve, and UI scripts

tameware:
Per a suggestion from @tzimnoch 

## Summary
- Extract PBN/LIN/DLM/sol parsers into `dds_web_deal_import.js` (#382)
- Split remaining glue into `dds_web_core.js` (deal model), `dds_web_solve.js` (WASM queue / DD table / leads), and `dds_web.js` (DOM UI)
- Wire HTML, GitHub Pages staging, Bazel data, and Node tests for the new classic-script load order

## Test plan
- [ ] `bazelisk test //web:dds_web_js_test //web:dds_web_html_test //web:stage_github_pages_test`
- [ ] Smoke `node --test web/tests/dds_web_test.mjs` with the four JS env paths set
- [ ] Open DDS Web locally and confirm import + sample deal + solve still work


web/tests/dds_web_test.mjs
```diff
@@ -226,8 +242,56 @@ function createMockDocument(initialValues = {}) {
...
+    const sandbox = {
+        console,
+        Promise,
+        Error,
+        ...extras,
+    };
+    const context = createContext(sandbox);
+    runInContext(code, context, { filename: "dds_web_deal_import.js" });
+    return context;
+}
```
tzimnoch, line 256: This is only used by a test. For deployed scenarios, runDdsWebScripts is the function called to load the scripts.

The tests should be updated to verify the behavior of runDdsWebScripts and these individual load functions can go away.

web/tests/dds_web_test.mjs
```diff
@@ -31,15 +31,31 @@ function findDdsWebJsPath() {
...
         }
     }
 
-    throw new Error("dds_web.js not found");
+    throw new Error(`${fileName} not found`);
+}
+
+function findDdsWebJsPath() {
+    return findWebJsPath("dds_web.js", "DDS_WEB_JS");
+}
```
tzimnoch, line 47: This is a style question, so feel free to ignore.

I'm not a fan of one-line refactors. If you wanted to know/verify the filename being loaded you have to click/scroll through to this function to verify.

For this sort of refactor to pay dividends the function needs to be called from different places, so that you only have to maintain the filename in one place.

When you remove the redundant loader that is only being employed for a test then there's only one caller of this function and the caller uses the filename themselves. It could be nice to see it back-to-back to verify they're the same.

Not a hill worth dying on. This is purely an aesthetic matter.

web/dds_web_core.js
```diff
@@ -0,0 +1,370 @@
...
+/* eslint-env es6 */
+/* exported DIRECTIONS SUITS PIPS DENOMINATIONS DENOM_TO_STRAIN DIR_TO_HAND
+            Card handsToPbn openingLeader pipFromDdsRank leadTricksMapFromSolverOutput
+            fourthHandFillState cardsToSuitHoldings sortedPipInsertIndex
+            allHandsHaveThirteenCards sanitizeSuitHolding
+            suitHoldingHasDuplicatePips suitHoldingHasIllegalChars */
+
+"use strict";
+
+(function (global) {
```
tzimnoch, line 19: 👍 ++

Took me a while to appreciate this pattern.

web/dds_web_core.js
```diff
@@ -0,0 +1,370 @@
...
+    const DIRECTIONS = ["north", "east", "south", "west"];
+    const SUITS = ["spades", "hearts", "diamonds", "clubs"];
+    const PIPS = "AKQJT98765432";
+    const DENOMINATIONS = ["C", "D", "H", "S", "N"];
```
tzimnoch, line 23: Some of this is repeated in dds_web_deal_import. Could be DRYed up. Not urgent.

tzimnoch (approved):
I would merge as-is and take (or leave) any comments as future suggestions.

tameware:
Will do! I’ll either address the comments shortly or make an issue for
them.

